PR #86, opened 2020-02-21
build: enable semantic-pull-request config

vinayakkulkarni:
- [x] Build-related changes

- [x] No

- [x] All tests are passing

**Other information:**
- [x] Enables semantic-pull-request support


posva:
Thank you but this doesn't really align with how we handle PRs. I don't think we need this since we squash commits in PRs so we don't really care about these options. I also want to limit this kind of build/ci related things and keep them as similar as possible to vue-next

PR #88, opened 2020-02-21
feat: add commitlint and husky

vinayakkulkarni:
- [x] Build-related changes

- [x] No

- [x] All tests are passing

**Other information:**
- [x] Ensures that the code pushed has no linting errors


posva:
Thank you but I think it's too early to add this and I prefer not having it

PR #89, opened 2020-02-22
fix #85 :  Reorder generic types on createStore function

### LINKED ISSUE #85: Permit to specify the State structure when using createStore
## Feature request
_[According to this twitter discussion](https://github.com/posva/pinia/blob/master/src/store.ts#L164)_
### Need
In order to getting more solid store typing, it could be nice to specify the type which will reflect the store. Currently, the state is inferred by the state initialization.
To be honest, it's already possible by this way : 
```ts
type MainStare = {
  counter: number;
  name: string;
}

export const useMainStore = createStore<string, MainState>({
  id: 'main',
  state: () => ({
    counter: 0,
    name: 'Edimitchel',
  }),
  // ...
})
```
As you can see, the current API forced us to specify the first generic type ([see here](https://github.com/posva/pinia/blob/master/src/store.ts#L164)).

### My request
**My simple request is following :** Could it be possible to swap [these two lines](https://github.com/posva/pinia/blob/master/src/store.ts#L163-L164) ?
_If accepted, I would create the PR_

The result would be like this (with a better usage if we have nested or optional state) : 

```ts
export type MainStare = {
  counter: number;
  name: string;
  team?: {
    name: string;
    createdAt?: Date;
  }
}

export const useMainStore = createStore<MainState>({
  id: 'main',
  state: () => ({
    counter: 0,
    name: 'Edimitchel',
    team: {
      name: 'Vue Community'
    }
  }),
  // ...
})
```
_______
I guess, following could be subject to another issue by amending the `createStore` signature :

```ts
export type MainStare = {
  counter: number;
  name: string;
  team?: {
    name: string;
    createdAt?: Date;
  }
}

export const useMainStore = createStore<MainState>('main', {
  state: () => ({
    counter: 0,
    name: 'Edimitchel',
    team: {
      name: 'Vue Community'
    }
  }),
  // ...
})
```

edimitchel:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)
- [x] All tests are passing
- [x] New/updated tests are included

- [ ] Making state more solid to ensure type safe !

See #85

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -161,9 +161,9 @@ export function buildStore<
  */
 export function createStore<
   S extends StateTree,
-  Id extends string = string,
   G extends Record<string, StoreGetter<S>> = Record<string, StoreGetter<S>>,
-  A extends Record<string, StoreAction> = Record<string, StoreAction>
+  A extends Record<string, StoreAction> = Record<string, StoreAction>,
+  Id extends string = string
 >(options: {
   id: Id
   state?: () => S
--- a/src/store.ts
+++ b/src/store.ts
@@ -1,4 +1,4 @@
-import { ref, watch, computed, reactive, Ref } from '@vue/composition-api'
+import { ref, watch, computed, Ref } from '@vue/composition-api'
 import {
   StateTree,
   StoreWithState,
@@ -160,10 +160,10 @@ export function buildStore<
  * @param options
  */
 export function createStore<
-  Id extends string,
   S extends StateTree,
-  G extends Record<string, StoreGetter<S>>,
-  A extends Record<string, StoreAction>
+  Id extends string = string,
+  G extends Record<string, StoreGetter<S>> = Record<string, StoreGetter<S>>,
+  A extends Record<string, StoreAction> = Record<string, StoreAction>
 >(options: {
   id: Id
   state?: () => S
--- a/__tests__/store.spec.ts
+++ b/__tests__/store.spec.ts
@@ -148,4 +148,43 @@ describe('Store', () => {
       store.state
     )
   })
+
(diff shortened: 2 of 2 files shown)


src/store.ts
```diff
@@ -160,10 +160,10 @@ export function buildStore<
  * @param options
  */
 export function createStore<
-  Id extends string,
   S extends StateTree,
-  G extends Record<string, StoreGetter<S>>,
-  A extends Record<string, StoreAction>
+  Id extends string = string,
```
posva: let's move it to the end

src/store.ts
```diff
@@ -160,10 +160,10 @@ export function buildStore<
...
+  G extends Record<string, StoreGetter<S>> = Record<string, StoreGetter<S>>,
+  A extends Record<string, StoreAction> = Record<string, StoreAction>
```
posva, line 166:
```suggestion
  G extends Record<string, StoreGetter<S>>,
  A extends Record<string, StoreAction>,
  Id extends string
```

posva: What I don't really like about the defaults is that you loose all typing if you only provide the State type in the generic `createStore<State>()`

I think it makes more sense to manually type the return of the function like said at https://github.com/posva/pinia/issues/85#issuecomment-631473019

edimitchel: Yes, it's a second way to do. But personally, I prefer to use generic feature and not polluating the body of my function.

PR #170, opened 2020-05-25
fix: avoid nuxtContext in spa mode

MartinMalinda:
Fixes https://github.com/posva/pinia/issues/161

Not calling the `setStateProvider()` in case of missing nuxtContext seems like a safer approach, otherwise it errored out on undefined state provider in some next step.

### DIFF
--- a/nuxt/plugin.js
+++ b/nuxt/plugin.js
@@ -16,7 +16,7 @@ const myPlugin = context => {
     context.beforeNuxtRender(({ nuxtState }) => {
       nuxtState.pinia = getRootState(context.req)
     })
-  } else {
+  } else if (context.nuxtState) {
     setStateProvider(() => context.nuxtState.pinia)
   }
 }


posva:
did you gave this a try in both universal and spa modes in nuxtconfig?

MartinMalinda:
Without this fix: `Cannot read property 'pinia' of undefined (in SPA mode)`

With this fix: 

```
[vue-composition-api] must call Vue.use(plugin) before using any function.
```

both (SPA and universal)

I guess this is becuase I tested via `npm link`. I'm not sure if I can test via pointing to github because `dist` is gitignored.

posva:
to test locally with linking, you need to link both `pinia` and `@vue/composition-api` in your app as well as `@vue/composition-api` in `pinia` to make sure both pinia and your App use the same library. It's a bit... convoluted

MartinMalinda:
@posva 

OK, so I just retested with a manual change in `node_modules` and it seems to be working fine both in `SPA` and `Universal`

posva:
Thanks!

PR #194, opened 2020-06-27
fix #193

### LINKED ISSUE #193: Cannot stringify POJOs with symbolic keys
When using Pinia from a fetch function in page with Nuxt, we get the following error for each Pinia store initiated.

```
WARN: Cannot stringify POJOs with symbolic keys Symbol(vfa.key.reactiveIdentifier)
```

It doesn't seem to affect the functionality, but this seems to be something that shouldn't happen...

Bug can be reproduced with a clean Nuxt installation:

```
npx create-nuxt-app piniabug
cd piniabug
yarn add pinia @vue/composition-api
# create a store according to example in Pinia README to stores/main.js
# setup the Nuxt plugin according to Pinia README
# setup composition-api as Nuxt plugin (see below)
# create a fetch function in the pages/index.vue file (see below)
```

### plugins/composition-api.js

```
import Vue from 'vue';
import VueCompositionApi from '@vue/composition-api';

Vue.use(VueCompositionApi);

export * from '@vue/composition-api';
```

### pages/index.vue (relevant parts)

```
import { useMainStore } from '@/stores/main'
...
 fetch({ req }) {
    const main = useMainStore(req);
    console.log(main.state);
  },
...
```

The error goes away if I use for example lodash cloneDeep on the req before passing it in to useMainStore, but that seems to break Pinia somehow, so that store modifications done during the fetch method are not available in the client-side.

nles:
Fixes https://github.com/posva/pinia/issues/193

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number)
- [ ] All tests are passing

**Other information:**

### DIFF
--- a/nuxt/plugin.js
+++ b/nuxt/plugin.js
@@ -14,7 +14,7 @@ const myPlugin = context => {
   if (process.server) {
     setActiveReq(context.req)
     context.beforeNuxtRender(({ nuxtState }) => {
-      nuxtState.pinia = getRootState(context.req)
+      nuxtState.pinia = JSON.parse(JSON.stringify(getRootState(context.req)));
     })
   } else if (context.nuxtState) {
     setStateProvider(() => context.nuxtState.pinia)


posva:
Thanks for the PR! See https://github.com/posva/pinia/issues/193#issuecomment-789673986

That being said we might want to transform to JSON the state for better perf but things will also be updated in docs. #378

PR #248, opened 2020-10-02
fix: only set state provider if a state exists

MartinMalinda:
Fixes https://github.com/posva/pinia/issues/206

At some point, Nuxt started to have `context.nuxtState` even in SPA mode so the previous check was not enough.
Maybe there's a smarter way to check for SPA mode of Nuxt, but I'm not aware of it.

### DIFF
--- a/nuxt/plugin.js
+++ b/nuxt/plugin.js
@@ -16,7 +16,7 @@ const myPlugin = context => {
     context.beforeNuxtRender(({ nuxtState }) => {
       nuxtState.pinia = getRootState(context.req)
     })
-  } else if (context.nuxtState) {
+  } else if (context.nuxtState && context.nuxtState.pinia) {
     setStateProvider(() => context.nuxtState.pinia)
   }
 }


posva:
but this could also fail as the state might not be set at the moment the plugin is installed, no? Is this an spa mode error only? There is no need to install the plugin in SPA mode as there is no hydration

MartinMalinda:
I think it's just a reiteration of the same error that was previously fixed: https://github.com/posva/pinia/pull/170

The previous check solved the problem efficiently, it's just that now it has to be more strict.

Yeah it is SPA mode only. Good point, I guess I could remove pinia from the buildPack to make it work in SPA. I sometimes run the app in SPA mode and sometimes in universal though, depending what I need, to have as fast rebuilds as possible. It's just a different CLI command for me. I don't know how common that is. But yeah I could use the same environment variable I use for `mode` and use it to either add the buildPack or not.

So perhaps I could remove the `else if` completely in this PR and add a note to readme.

posva:
No, I think the direction of the PR is good, it should be simple and straightforward, it doesn't even add to the client bundle. But we can maybe read the mode or check if the app was server rendered. I'm pretty sure there must be something in the context that says that

MartinMalinda:
I'm looking into the `context` and I can't find anything. There is `isStatic` which probably refers to `mode: static` but that's a different thing. There's no `isSPA`.

Only explicit check I can think of is `!!document.querySelector('#__nuxt > #nuxt-loading')` which is probably even more hacky:)

MartinMalinda:
I take it back. There is now `context.nuxtState.isServerRendered` which could be used. It could crash at some point in the future if `nuxtState` becomes undefined again in SPA mode... but I guess that's not gonna happen.

Although `isServerRendered` is not documented and google does not help much. Not sure if it's a reliable flag.

posva:
~~I checked and it's okay to use `isServerRendered`!~~

Actually having a simple check like you did is easier and better

posva (approved):

PR #334, opened 2021-01-15
fix: resilient _VUE_DEVTOOLS_TOAST_ 

timozander:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [x] No

- [x] All tests are passing

**Other information:**
When using Pinia v2 with the latest beta of Vue devtools, I always experience the following console error:

```
backend.js:3166 ReferenceError: __VUE_DEVTOOLS_TOAST__ is not defined
    at pinia.esm-bundler.js?v=94b0b286:232
    at Object.addPlugin (backend.js:3162)
    at Object.<anonymous> (backend.js:908)
    at Object.emit (<anonymous>:1:780)
    at Object._replayBuffer (<anonymous>:1:332)
    at Object.on (<anonymous>:1:424)
    at connect (backend.js:907)
```

While I cannot exactly pin down the exact cause of this, I think its reasonable to fall back to console output whenever the devtools's toast is not available

### DIFF
--- a/src/devtools.ts
+++ b/src/devtools.ts
@@ -19,9 +19,9 @@ function toastMessage(
   message: string,
   type?: 'normal' | 'error' | 'warning' | undefined
 ) {
-  const piniaMessage = '🍍' + message
+  const piniaMessage = '🍍 ' + message
 
-  if (typeof __VUE_DEVTOOLS_TOAST__ !== 'undefined') {
+  if (typeof __VUE_DEVTOOLS_TOAST__ === 'function') {
     __VUE_DEVTOOLS_TOAST__(piniaMessage, type)
   } else if (type === 'error') {
     console.error(piniaMessage)
--- a/src/devtools.ts
+++ b/src/devtools.ts
@@ -15,21 +15,20 @@ function formatDisplay(display: string) {
   }
 }
 
-function outputMessage(
+function toastMessage(
   message: string,
   type?: 'normal' | 'error' | 'warning' | undefined
 ) {
-  if (__VUE_DEVTOOLS_TOAST__) {
-    __VUE_DEVTOOLS_TOAST__(message, type)
-    return
-  }
+  const piniaMessage = '🍍' + message
 
-  if (type === 'error') {
-    console.error(message)
+  if (typeof __VUE_DEVTOOLS_TOAST__ !== 'undefined') {
+    __VUE_DEVTOOLS_TOAST__(piniaMessage, type)
+  } else if (type === 'error') {
+    console.error(piniaMessage)
   } else if (type === 'warning') {
-    console.warn(message)
+    console.warn(piniaMessage)
   } else {
-    console.log(message)
+    console.log(piniaMessage)
   }
 }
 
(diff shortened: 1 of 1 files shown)


posva:
This is weird since the devtools should be installed at that point. I would need to have more details to ensure this is the right fix

timozander:
I found out that it is caused by the devtools failing to load a specific source (https://github.com/vuejs/vue-devtools/issues/1322#issuecomment-752352360)

However, I think that adding this extra sanity check is reasonable since relying on globally available methods like this is always at the risk of failure, causing console spam

src/devtools.ts
```diff
@@ -15,6 +15,24 @@ function formatDisplay(display: string) {
   }
 }
 
+function outputMessage(
```
posva: let's call it `toastMessage`

src/devtools.ts
```diff
@@ -15,6 +15,24 @@ function formatDisplay(display: string) {
...
+  message: string,
+  type?: 'normal' | 'error' | 'warning' | undefined
+) {
+  if (__VUE_DEVTOOLS_TOAST__) {
```
posva:
```suggestion
  if (typeof __VUE_DEVTOOLS_TOAST__ !== undefined) {
```

src/devtools.ts
```diff
@@ -15,6 +15,24 @@ function formatDisplay(display: string) {
...
+    __VUE_DEVTOOLS_TOAST__(message, type)
+    return
```
posva: remove this return and use an `else`

src/devtools.ts
```diff
@@ -116,18 +134,15 @@ export function addDevtools(app: App, store: GenericStore) {
               options: formatStoreForInspectorState(store),
             }
           } else {
-            __VUE_DEVTOOLS_TOAST__(
-              `🍍 store "${payload.nodeId}" not found`,
-              'error'
-            )
+            outputMessage(`🍍 store "${payload.nodeId}" not found`, 'error')
```
posva: let's move the `🍍 ` inside of the `toastMessage` in front of every message

posva (changes requested):

src/devtools.ts
```diff
@@ -15,6 +15,23 @@ function formatDisplay(display: string) {
   }
 }
 
+function toastMessage(
+  message: string,
+  type?: 'normal' | 'error' | 'warning' | undefined
+) {
+  const piniaMessage = '🍍' + message
```
posva:
```suggestion
  const piniaMessage = '🍍 ' + message
```

src/devtools.ts
```diff
@@ -15,6 +15,23 @@ function formatDisplay(display: string) {
...
+
+  if (typeof __VUE_DEVTOOLS_TOAST__ !== 'undefined') {
```
posva:
```suggestion
  if (typeof __VUE_DEVTOOLS_TOAST__ === 'function') {
```

posva (approved):
Thanks!

PR #444, opened 2021-04-18
feat(devtools): work with stores added before app.use

marshallswain:
This removes the need to call `app.use(pinia)` BEFORE registering all services.  Services registered before calling the plugin are now held by a promise until the `app` object is set by the plugin.  The code used to set and get the `app` object for the devtools to work has been updated.  

As noted in the checkboxes, below, I haven't added any tests for this.  I did copy the built version into my Vite app to test. The app I'm testing has three stores that register before the app is created and five that register afterward.  Here's what I noticed from these changes:

- [x] Stores registered both before and after `app.use(pinia)` display properly in the devtools.
- [x] No extra logging occurs
- [x] No errors occur
- [x] Each store's data is browseable.

Summary of changes:
- convert getClientApp() to return a promise.  The promise resolves when the app is set (and a boolean ref gets set to true)
- Removes logging and warning state related to registering stores before calling app.use(pinia)

- [x] Feature

- [x] No

- [x] When resolving a specific issue, it's referenced in the PR's title (e.g. `fix #xxx[,#xxx]`, where "xxx" is the issue number) **not applicable**
- [x] All tests are passing

- [x] A convincing reason for adding this feature: Nobody will ask you for this feature in the future.  ;)

### DIFF
--- a/src/rootStore.ts
+++ b/src/rootStore.ts
@@ -1,4 +1,4 @@
-import { App, InjectionKey, Plugin, Ref, ref, warn } from 'vue'
+import { App, InjectionKey, Plugin, Ref, ref, warn, watch } from 'vue'
 import { IS_CLIENT } from './env'
 import {
   StateTree,
@@ -53,8 +53,23 @@ export const storesMap = new WeakMap<
  * Client-side application instance used for devtools
  */
 export let clientApp: App | undefined /*#__PURE__*/
-export const setClientApp = (app: App) => (clientApp = app)
-export const getClientApp = () => clientApp
+export const isClientAppReady = ref(false)
+export const setClientApp = (app: App) => {
+  clientApp = app
+  isClientAppReady.value = true
+}
+let promise: Promise<App> | undefined
+export const getClientApp = () => {
+  promise =
+    promise ||
+    new Promise((resolve) => {
+      watch(
+        () => isClientAppReady.value,
+        (val: boolean) => val && resolve(clientApp as App)
+      )
+    })
+  return promise
+}
 
 /**
  * Plugin to extend every store
--- a/src/rootStore.ts
+++ b/src/rootStore.ts
@@ -1,4 +1,4 @@
-import { App, InjectionKey, Plugin, Ref, ref, warn, watch } from 'vue'
+import { App, InjectionKey, Plugin, Ref, ref, warn } from 'vue'
 import { IS_CLIENT } from './env'
 import {
   StateTree,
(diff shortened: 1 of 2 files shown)


marshallswain:
@posva maybe I should have checked with you before making this PR.  Not sure if you've seen it, yet.  I know conference time is busier than normal.  Does this look like something worth merging?

src/rootStore.ts
```diff
@@ -53,8 +53,23 @@ export const storesMap = new WeakMap<
...
+export const setClientApp = (app: App) => {
+  clientApp = app
+  isClientAppReady.value = true
+}
+let promise: Promise<App> | undefined
+export const getClientApp = () => {
+  promise =
+    promise ||
+    new Promise((resolve) => {
+      watch(
```
posva: rather than creating a watcher, I think we should keep the resolve function in a variable that is then called by `setClientApp()`

src/rootStore.ts
```diff
@@ -53,8 +53,23 @@ export const storesMap = new WeakMap<
...
-export const setClientApp = (app: App) => (clientApp = app)
-export const getClientApp = () => clientApp
+export const isClientAppReady = ref(false)
+export const setClientApp = (app: App) => {
+  clientApp = app
+  isClientAppReady.value = true
+}
+let promise: Promise<App> | undefined
+export const getClientApp = () => {
+  promise =
```
posva: I think this should rather be

```js
return clientApp ? Promise.resolve(clientApp) : new Promise(...)
```

(also probably without the return)

posva (changes requested):
Thanks for this!

src/rootStore.ts
marshallswain: Wow.  I've literally never used that pattern, but that's way easier.  Thanks!

marshallswain:
@posva With your suggestions I was able to make it even more succinct.  Note that I also removed the exported `clientApp` variable, since we can just pull it from the promise.  I noticed that the devtools still work without it, which I figured was the primary concern for removing it.

marshallswain:
I also took a look at the docs and it doesn't seem like anything needs to be updated.

marshallswain:
I also took a look at the docs and it doesn't seem like anything needs to be updated.

src/rootStore.ts
```diff
@@ -50,11 +50,14 @@ export const storesMap = new WeakMap<
 >()
 
 /**
- * Client-side application instance used for devtools
+ * Expose the client-side application instance used for devtools
  */
-export let clientApp: App | undefined /*#__PURE__*/
-export const setClientApp = (app: App) => (clientApp = app)
-export const getClientApp = () => clientApp
+let promise: Promise<App> | undefined
```
posva:
```suggestion
let clientAppPromise: Promise<App> | undefined
```

src/rootStore.ts
```diff
@@ -50,11 +50,14 @@ export const storesMap = new WeakMap<
...
+let resolveApp: Function | undefined
```
posva:
```suggestion
let resolveApp: ((app: App) => void) | undefined
```

src/rootStore.ts
```diff
@@ -50,11 +50,14 @@ export const storesMap = new WeakMap<
...
+export const setClientApp = (app: App) =>
+  resolveApp && (resolveApp as Function)(app)
```
posva: with the change above the cast should be able to disappear
```suggestion
  resolveApp && resolveApp(app)
```

posva (changes requested):
Nice, looking good

marshallswain:
@posva thanks for the feedback.  I've implemented the changes.

posva (approved):
Thanks!

PR #597, opened 2021-08-04
feat: destroy a store with $dispose

### LINKED ISSUE #557: Add ability to destroy stores

Sometimes we need to define stores dynamically - e.g. to be able to use 2 instances of the same store that don't share the same state. 

In such a case, in Vuex, I create a factory function that takes dependencies and return an initialized instance of a Vuex store. Then, when I no longer need that store I can call `unregisterModule()` and destroy that store. 

In Pinia I didn't find such an option.

### Proposed solution

``` vue
<script>
import createMyStore from '@/store/createMyStore';
import { onUnmounted } from '@vue/composition-api';

export default {
  name: 'App',

  setup() {
    const useStore = createMyStore({
      id: 'myId',
      useSettings: () => ({
        value: 1,
      })
    });
    const { combinedValue, $destroy } = useStore();

    onUnmounted(() => {
      $destroy();
    })

    return {
      combinedValue,
    }
  },
}
</script>
```

``` js
import { defineStore } from 'pinia'

export default ({ useSettings, id }) => {
  const mainId = 'myStore'

  return defineStore({
    id: `${mainId}_${id}`,

    state: () => ({
      value: 1,
    }),

    getters: {
      combinedValue() {
        const settings = useSettings();
        return this.value + settings.value;
      },
    },
  })
}
```

I don't think there's any alternative. We can call `$reset()` to restore the state - this will free-up the memory, but it won't destroy the instance entirely.

posva:
Close #557

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

- [ ] All tests are passing

**Other information:**

### DIFF
--- a/packages/pinia/__tests__/store.spec.ts
+++ b/packages/pinia/__tests__/store.spec.ts
@@ -303,4 +303,27 @@ describe('Store', () => {
     expect(One.text()).toBe('1')
     expect(Two.text()).toBe('1')
   })
+
+  it('can be disposed', () => {
+    const pinia = createPinia()
+    const useStore = defineStore({
+      id: 'main',
+      state: () => ({ n: 0 }),
+    })
+
+    const store = useStore(pinia)
+    const spy = jest.fn()
+
+    store.$subscribe(spy)
+    pinia.state.value.main.n++
+    expect(spy).toHaveBeenCalledTimes(1)
+
+    expect(useStore()).toBe(store)
+    store.$dispose()
+    pinia.state.value.main.n++
+
+    expect(spy).toHaveBeenCalledTimes(1)
+
+    expect(useStore()).not.toBe(store)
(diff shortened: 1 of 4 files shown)


sabifa:
How am I supposed to use `disposePinia`? It seems like I can't import it in any way.

eastmann:
> How am I supposed to use `disposePinia`? It seems like I can't import it in any way.

It seems you should distructure `$dispose` from your **store instance**:

```
import { onUnmounted } from 'vue'
import { useYourStore } from '@/stores/YourStore'
...
const { $dispose } = useYourStore()
...
onUnmounted(() => {
    $dispose()
})
```

PR #621, opened 2021-08-21
docs: add store destructure guide

TomatoGuy0502:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Build-related changes
- [x] Other, please describe: Docs update

- [ ] Yes
- [x] No

- [ ] All tests are passing

**Other information:**

I found that there is a method `storeToRefs` in the source code, but didn't saw it in docs.
So I think it's good to put it in store guide.

### DIFF
--- a/packages/docs/core-concepts/index.md
+++ b/packages/docs/core-concepts/index.md
@@ -39,7 +39,7 @@ If you are not using `setup` components yet, [you can still use Pinia with _map 
 
 Once the store is instantiated, you can access any property defined in `state`, `getters`, and `actions` directly on the store. We will see these in detail in the next pages but autocompletion will help you.
 
-Note that `store` is an object wrapped with `reactive`, meaning there is no need to write `.value` after getters but, like `props` in `setup`, **we cannot destructure it**:
+Note that `store` is an object wrapped with `reactive`, meaning there is no need to write `.value` after getters but, like `props` in `setup`, **we cannot destructure it directly**:
 
 ```js
 export default defineComponent({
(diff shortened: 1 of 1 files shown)


packages/docs/core-concepts/index.md
```diff
@@ -39,7 +39,7 @@ If you are not using `setup` components yet, [you can still use Pinia with _map
 
 Once the store is instantiated, you can access any property defined in `state`, `getters`, and `actions` directly on the store. We will see these in detail in the next pages but autocompletion will help you.
 
-Note that `store` is an object wrapped with `reactive`, meaning there is no need to write `.value` after getters but, like `props` in `setup`, **we cannot destructure it**:
+Note that `store` is an object wrapped with `reactive`, meaning there is no need to write `.value` after getters but, like `props` in `setup`, **we cannot destructure it directly**:
```
posva: Let's keep it the way it was

```suggestion
Note that `store` is an object wrapped with `reactive`, meaning there is no need to write `.value` after getters but, like `props` in `setup`, **we cannot destructure it**:
```

packages/docs/core-concepts/index.md
```diff
@@ -63,3 +63,22 @@ export default defineComponent({
...
 ```
+
+You can use `storeToRefs` to destructure the returned object without losing reactivity:
+```js
+import { storeToRefs } from 'pinia'
+
+export default defineComponent({
+  setup() {
+    const store = useStore()
+    // Now name & doubleCount still have reactivity
```
```suggestion
    // `name` and `doubleCount` are reactive refs
```

packages/docs/core-concepts/index.md
```diff
@@ -63,3 +63,22 @@ export default defineComponent({
...
+    // Note that methods and non reactive properties will be ignored
```
```suggestion
    // This will also create refs for properties added by plugins
    // but skip any action or non reactive (non ref/reactive) property
```

packages/docs/core-concepts/index.md
```diff
@@ -63,3 +63,22 @@ export default defineComponent({
   },
 })
 ```
+
+You can use `storeToRefs` to destructure the returned object without losing reactivity:
```
```suggestion
In order to extract properties from the store while keeping its reactivity, you need to use `storeToRefs()`. It will create refs for any reactive property. This is useful when you are only using state from the store but not calling any action:
```

posva (changes requested):
Thanks, I added a few comments. Let me know what you think

TomatoGuy0502 (commented):
Yes, I think that's much more clear!

PR #633, opened 2021-08-25
docs: add getters method-style Access example

reslear:
from https://github.com/posva/pinia/issues/342

### DIFF
--- a/packages/docs/core-concepts/getters.md
+++ b/packages/docs/core-concepts/getters.md
@@ -79,6 +79,36 @@ export const useStore = defineStore('main', {
 })
 ```
 
+
+## Accessing Method-Style getters
+
+Use if you need pass arguments:
+
+```js
+export const useStore = defineStore('main', {
+  getters: {
+    getUserById() {
+      return (userId) => this.users.find((user) => user.id === userId)
+    },
+  },
+})
+```
+
+and use in component:
+
+```vue
+<script>
+export default {
+  setup() {
+    const store = useStore()
+    const getUserById = (id) => store.getUserById(id)
+
+    return { getUserById }
+  },
+}
+</script>
+```
+
 ## Accessing other stores getters
 
 To use another store getters, you can directly _use it_ inside of the _action_:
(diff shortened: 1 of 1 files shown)


packages/docs/core-concepts/getters.md
```diff
@@ -79,6 +79,36 @@ export const useStore = defineStore('main', {
 })
 ```
 
+
+## Accessing Method-Style getters
+
+Use if you need pass arguments:
```
```suggestion
## Passing arguments to getters

_Getters_ are just _computed_ properties behind the scenes, so it's not possible to pass any parameters to them. However, you can return a function from the _getter_ to accept any arguments:
```

packages/docs/core-concepts/getters.md
```diff
@@ -79,6 +79,36 @@ export const useStore = defineStore('main', {
...
+
+```js
+export const useStore = defineStore('main', {
+  getters: {
+    getUserById() {
+      return (userId) => this.users.find((user) => user.id === userId)
+    },
```
```suggestion
    getUserById: (state) => {
      return (userId) => state.users.find((user) => user.id === userId)
    },
```

packages/docs/core-concepts/getters.md
```diff
@@ -79,6 +79,36 @@ export const useStore = defineStore('main', {
...
+<script>
+export default {
+  setup() {
+    const store = useStore()
+    const getUserById = (id) => store.getUserById(id)
+
+    return { getUserById }
+  },
+}
+</script>
```
```suggestion
<script>
export default {
  setup() {
    const store = useStore()

    return { getUserById: store.getUserById }
  },
}
</script>

<template>
User 2: {{ getUserById(2) }}
</template>
```

packages/docs/core-concepts/getters.md
```diff
@@ -79,6 +79,36 @@ export const useStore = defineStore('main', {
...
+```
+
```
posva, line 127:
````suggestion

Note that when doing this, **getters are not cached anymore**, they are simply functions that you invoke. You can however cache some results inside of the getter itself, which is uncommon but should prove more performant:

```js
export const useStore = defineStore('main', {
  getters: {
    getActiveUserById(state) {
      const activeUsers = state.users.filter((user) => user.active)
      return (userId) => activeUsers.find((user) => user.id === userId)
    },
  },
})
```

````

posva (approved):
Thanks!

PR #646, opened 2021-08-31
docs: Add notes on mapState and undefined properties. #641

rbanks54:
Documentation updates as per discussion #641


posva:
Thank you for this! I was more thinking of adding comments to the source code (like one of the changes you did). They are now generated as part of the docs: https://pinia.esm.dev/api/interfaces/pinia.DefineStoreOptions.html#state

That being said, a dev only warning could be more helpful in this scenario:

```js
if (__DEV__ && b && typeof b === 'object' && typeof b.constructor === 'function' && b.constructor.toString().includes('[native code]')) {
 warn here
}
```

rbanks54:
I'll see if I can get the warning working. If I get it working, I'll adjust the docs as well

rbanks54:
Hmm.... looking into it further, I'm not going to be able to easily produce the warning 🙁

It requires runtime checks after the store and state objects are instantiated, and I'm hesitant to add too much overhead to the mapState methods due to potential performance impacts.

For example, I can check array mappers by doing something like this:
```
//...mapState helper code
  return Array.isArray(keysOrMapper)
    ? keysOrMapper.reduce((reduced, key) => {
      reduced[key] = function (this: ComponentPublicInstance) {
        const store = useStore(this.$pinia)
        warnIfPropertyNotPresent(store, key); //  <----- Add this method call
        return useStore(this.$pinia)[key]

//....skip

const warnIfPropertyNotPresent = (store: any, keyName: any) => {
  // This is slower that hasOwnProperty, but will still succeed when property is from super class
  if (__DEV__ && !(keyName in store.$state)) {
    console.warn(`mapState: property '${keyName}' is not present in the store's state object. Mapping will not be reactive.`)
  }
}
```

It's workable, but adds an overhead per call that probably doesn't pay for itself in terms of usefulness compared with the performance impact.

Unfortunately, it doesn't help when when we use the function mapper approach. When it comes to the function call, I can't tell if the function references a property that isn't present, or if the property exists and just has an undefined value.

The `.call()` method on line 274 below returns the result of the function with no easy way to inspect the function itself to see what properties are being referenced.

https://github.com/posva/pinia/blob/9a4cc2ebdd90af5d2bab96e5b52618c6016fcf30/packages/pinia/src/mapHelpers.ts#L273-L276

For reference, internally the call ends up deep in Vue's reactivity system, eventually hitting this code in `createGetter()`

```
    const res = Reflect.get(target, key, receiver)
   //...
    if (isRef(res)) {
      // ref unwrapping - does not apply for Array + integer key.
      const shouldUnwrap = !targetIsArray || !isIntegerKey(key)
      return shouldUnwrap ? res.value : res
    }
```

[link](https://github.com/vuejs/vue-next/blob/47f488350c53db2bd9682e0c27a4f1dbcdab162a/packages/reactivity/src/baseHandlers.ts#L107-L125)

The `Reflect.get()` call will return a `ref` for any property that exists, and returns `undefined` for properties that don't exist. Sort of useful, but I don't think we really want to call `Reflect.get()` and do function parsing whenever the mapper function is called. Performance considerations, again.

Admittedly, Reflect would be useful to detect when we refer to a non-existent property key, but it's an internal detail within Vue's reactivity system and the ref is unwrapped as part of the method call in any case. All we get back in `mapState()` is `undefined`, regardless of if Reflect returns a ref() or not.

I think the docs approach is the simpler path, unless you think the warning c/should come from Vue itself?

posva:
Oh yeah, that's not what I was thinking at all 😓 What I was thinking was to warn when a store **is created** and its `state` function returns a non-plain object (like a class instance). It will warn when doing this

```ts
class MyState {
  someValue?: string;
}
const useMyStore = defineStore({
  state: () => new MyState()
  //...
}

useMyStore() // warns!
```

The changes shouldn't affect `mapState` nor the docs

rbanks54:
Heh. OK, so we'll do a more generalised "just be careful when doing this" style warning, rather than a "hey, this definitely won't work as you expected" warning.

rbanks54:
Updated the PR. Is this more what you're thinking of?

posva (changes requested):
This is getting closer but it's still not it: it should test that the return value of the setup function (in the createSetupStore) is non extended plain object (the snippet I passed above should do)

rbanks54:
Thanks for being patient. This sort has been a good way to pick up some new JS knowledge 🤔 😊

posva (changes requested):
Thanks but this is still inside of the same function 😅

The relevant code should be inside the createSetupStore, after the setup function is called, by doing a dev only check on the returned value. You shouldn't need to refactor anything

rbanks54:
Attempt number 3526 😁 

The previous change was to try and keep the warning as close to earliest detection point I could find, and looking at the code further I see why keeping it in the createSetupStore function is a little better. The length of that function and the named inline functions makes it quite hard to follow what's going on so I was losing where the function ends and thinking I was looking at a different scope.

P.S. I needed to merge in upstream changes due to a whitespace conflict git was convinced existed so there's a branch commit in the history. I'm hoping you'll squash this merge so that all the noise in it doesn't leak into your repo 🤔 😄

packages/pinia/src/store.ts
```diff
@@ -655,6 +655,19 @@ function createSetupStore<
...
+  if (
+    __DEV__ &&
+    store.$state &&
+    typeof store.$state === 'object' &&
+    typeof store.$state.constructor === 'function' &&
+    !store.$state.constructor.toString().includes('[native code]')
+  ) {
+    console.warn(
+      `[🍍]: Detected constructor usage in state initialisation.\n` +
+        `Ensure any undefined properties are explicitly initialised for reactivity to work as expected.`
```
```suggestion
      `[🍍]: The "state" must be a plain object. It cannot be\n\tstate: () => new MyClass()` 
```

packages/pinia/__tests__/store.spec.ts
```diff
@@ -323,4 +326,46 @@ describe('Store', () => {
 
     expect(useStore()).not.toBe(store)
   })
+
+  it('warns when state is created with a class constructor', () => {
+    class MyState {
+      someValue?: string
+    }
```
posva:
```suggestion
    class MyState {}
```

packages/pinia/__tests__/store.spec.ts
```diff
@@ -323,4 +326,46 @@ describe('Store', () => {
...
+
+    const useMyStore = defineStore({
+      id: 'functionInit',
+      state(): MyState {
+        return new MyState()
+      },
```
posva:
```suggestion
      state: () => new MyState(),
```

packages/pinia/__tests__/store.spec.ts
```diff
@@ -323,4 +326,46 @@ describe('Store', () => {
...
     expect(useStore()).not.toBe(store)
   })
+
+  it('warns when state is created with a class constructor', () => {
+    class MyState {
+      someValue?: string
+    }
+
+    const useMyStore = defineStore({
+      id: 'functionInit',
```
posva:
```suggestion
      id: 'store',
```

packages/pinia/__tests__/store.spec.ts
```diff
@@ -323,4 +326,46 @@ describe('Store', () => {
...
+    })
+    useMyStore()
+    expect('Detected constructor usage').toHaveBeenWarned()
+
+    const useMyOtherStore = defineStore({
+      id: 'arrowInit',
+      state: () => new MyState(),
+    })
+    useMyOtherStore()
+    expect('Detected constructor usage').toHaveBeenWarned()
```
posva:
```suggestion
```

packages/pinia/__tests__/store.spec.ts
```diff
@@ -323,4 +326,46 @@ describe('Store', () => {
...
+  })
+
+  it('only warns about constructors when store is initially created', () => {
+    class MyState { someValue?: string }
```
posva:
```suggestion
    class MyState {}
```

packages/pinia/__tests__/store.spec.ts
```diff
@@ -323,4 +326,46 @@ describe('Store', () => {
...
+    const useMyStore = defineStore({
+      id: 'arrowInit',
+      state: () => new MyState(),
+    });
+    useMyStore();
+    useMyStore();
```
posva:
```suggestion
```

packages/pinia/__tests__/store.spec.ts
```diff
@@ -323,4 +326,46 @@ describe('Store', () => {
...
+    expect('Detected constructor usage').toHaveBeenWarned()
+  })
+
+  it('only warns about constructors when store is initially created', () => {
+    class MyState { someValue?: string }
+    const useMyStore = defineStore({
+      id: 'arrowInit',
+      state: () => new MyState(),
+    });
+    useMyStore();
```
posva:
```suggestion
    useMyStore()
```

packages/pinia/__tests__/store.spec.ts
```diff
@@ -323,4 +326,46 @@ describe('Store', () => {
...
+    useMyStore();
+    expect('Detected constructor usage').toHaveBeenWarnedTimes(1);
```
posva:
```suggestion
    expect('Detected constructor usage').toHaveBeenWarnedTimes(1)
```

posva (approved):
Thanks and sorry for the delay!

PR #661, opened 2021-09-08
feat(docs): add action example in setup mode

coyotte508:
As a first time user, a lot of things that are now obvious were not to me. Like the fact that you don't need to call `this.$patch` inside actions in setup mode.

### DIFF
--- a/packages/docs/introduction.md
+++ b/packages/docs/introduction.md
@@ -55,8 +55,9 @@ You can even use a function (similar to a component `setup()`) to define a Store
 ```js
 export const useCounterStore = defineStore('counter', () => {
   const count = ref(0)
+  const increment = () => count.value++
 
-  return { count }
+  return { count, increment }
 })
 ```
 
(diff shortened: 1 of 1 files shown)


posva (changes requested):
Thanks! Could you add the action the the option example above, too?

packages/docs/introduction.md
```diff
@@ -55,8 +62,9 @@ You can even use a function (similar to a component `setup()`) to define a Store
 ```js
   const count = ref(0)
+  const increment = () => count.value++
```
posva:
```suggestion
  function increment() {
    count.value++
  }
```

packages/docs/introduction.md
```diff
@@ -46,6 +51,8 @@ export default {
     counter.count++
     // with autocompletion ✨
     counter.$patch({ count: counter.count + 1 })
+    // using the action
```
posva:
```suggestion
    // or using an action instead
```

posva (approved):
Thanks!

PR #683, opened 2021-09-17
Update code example in docs

filipsobol:


packages/docs/core-concepts/index.md
```diff
@@ -49,14 +49,14 @@ export default defineComponent({
...
-    doubleCount // 2
+    console.log(name); // "eduardo"
+    console.log(doubleCount); // 2
 
     return {
       // will always be "eduardo"
       name,
       // will always be 2
-      doubleCount
+      doubleCount,
```
posva, line 59: Let's keep only this line change

posva (changes requested):
Thanks!

packages/docs/core-concepts/index.md
filipsobol: Updated

posva (approved):

PR #697, opened 2021-10-03
docs: minor typo and grammar fixes

shasderias:
Just the really apparent ones.

### DIFF
--- a/packages/docs/core-concepts/actions.md
+++ b/packages/docs/core-concepts/actions.md
@@ -18,7 +18,7 @@ export const useStore = defineStore('main', {
 })
 ```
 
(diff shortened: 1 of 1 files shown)


packages/docs/core-concepts/actions.md
```diff
@@ -62,9 +62,9 @@ export default defineComponent({
 })
 ```
 
-## Accessing other stores actions
+## Accessing other stores' actions
 
-To use another store, you can directly _use it_ inside of the _action_:
+To use another store's action, you can directly _use it_ inside of the _action_:
```
posva: This one was correct

posva (changes requested):
Thanks!

packages/docs/core-concepts/actions.md
shasderias: Apologies.

Removed the change to the heading as well. The other headings in the docs are written the same way and I rather a PR to address them all. I'll attempt one if I find the time. Let's just do the really obvious typos in this one.

posva (approved):

PR #702, opened 2021-10-05
feat(warn): log store id with class constructor warning

shayneo:
We were using class constructors to create state in some of our stores, thought it would be helpful to include the store id in the log to that it would be easier to spot which stores need to be converted to plain objects.

### DIFF
--- a/packages/pinia/__tests__/store.spec.ts
+++ b/packages/pinia/__tests__/store.spec.ts
@@ -335,38 +335,35 @@ describe('Store', () => {
   })
 
   const warnTextCheckPlainObject = (storeId: string) =>
-    `"state" for ${storeId} must be a plain object`
+    `The "state" must be a plain object. It cannot be\n\tstate: () => new MyClass()\nFound in store "${storeId}".`
 
   it('warns when state is created with a class constructor', () => {
     class MyState {}
 
-    const id = 'store'
     const useMyStore = defineStore({
-      id,
+      id: 'store',
       state: () => new MyState(),
     })
     useMyStore()
-    expect(warnTextCheckPlainObject(id)).toHaveBeenWarned()
+    expect(warnTextCheckPlainObject('store')).toHaveBeenWarned()
   })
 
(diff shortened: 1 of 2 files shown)


packages/pinia/__tests__/store.spec.ts
```diff
@@ -334,35 +334,39 @@ describe('Store', () => {
...
-  const warnTextCheckPlainObject = `"state" must be a plain object`
+  const warnTextCheckPlainObject = (storeId: string) =>
+    `"state" for ${storeId} must be a plain object`
 
   it('warns when state is created with a class constructor', () => {
     class MyState {}
 
+    const id = 'store'
     const useMyStore = defineStore({
-      id: 'store',
```
posva: Let's keep this as a repeated string in tests, without creating the `const id = 'store'`

packages/pinia/src/store.ts
```diff
@@ -676,7 +676,7 @@ function createSetupStore<
     !store.$state.constructor.toString().includes('[native code]')
   ) {
     console.warn(
-      `[🍍]: The "state" must be a plain object. It cannot be\n\tstate: () => new MyClass()`
+      `[🍍]: The "state" for ${store.$id} must be a plain object. It cannot be\n\tstate: () => new MyClass()`
```
```suggestion
      `[🍍]: The "state" must be a plain object. It cannot be\n\tstate: () => new MyClass()\nFound in store "${store.id}".`
```

packages/pinia/__tests__/store.spec.ts
```diff
@@ -334,35 +334,39 @@ describe('Store', () => {
     expect(useStore()).not.toBe(store)
   })
 
-  const warnTextCheckPlainObject = `"state" must be a plain object`
+  const warnTextCheckPlainObject = (storeId: string) =>
+    `"state" for ${storeId} must be a plain object`
```
posva: This will need to be adapted too with the updated message

posva (changes requested):
Thanks, this is useful!

posva (approved):

PR #726, opened 2021-10-17
docs: Fix typo

Rolanddoda:


posva (changes requested):
Thanks! Let's also replace changing with mutating

posva (approved):

PR #745, opened 2021-10-26
fix(plugins): ensure plugins are used only once

### LINKED ISSUE #744: Plugins are duplicated for every createApp(...).use(pinia) call
When creating a new store, [all the plugins in the `_p` list are executed](https://github.com/posva/pinia/blob/pinia%402.0.0-rc.15/packages/pinia/src/store.ts#L648-L677). The `_p` list is filled in the `Pinia.install` function, by [pushing all the plugins from the `toBeInstalled` list into the _p list](https://github.com/posva/pinia/blob/pinia%402.0.0-rc.15/packages/pinia/src/createPinia.ts#L38).

This means that, if the install function is executed multiple times (e.g. by re-initializing the Vue app with the same Pinia instance), the registered plugins end up in the `_p` list multiple times as well, and will be executed repeatedly when a new store is created.

### Reproduction
```js
import { createApp } from 'vue';
import { defineStore, createPinia } from 'pinia';

const plugin = (context) => {
	context.store.count++;
};

const pinia = createPinia();
pinia.use(plugin);

const useStore = defineStore('my-store', {
	state: () => ({ count: 0 }),
});

createApp({}).use(pinia);
createApp({}).use(pinia);
createApp({}).use(pinia);

const store = useStore();
console.log(store.count); // Expected '1', but prints '3'
```

jessevanassen:
Fix #744

### DIFF
--- a/packages/pinia/__tests__/storePlugins.spec.ts
+++ b/packages/pinia/__tests__/storePlugins.spec.ts
@@ -251,10 +251,8 @@ describe('store plugins', () => {
   it('only executes plugins once after multiple installs', async () => {
     const pinia = createPinia()
 
-    let invokedCount = 0
-    pinia.use(() => {
-      invokedCount++
-    })
+    const spy = jest.fn()
+    pinia.use(spy)
 
     for (let i = 0; i < 3; i++) {
       mount({ template: 'none' }, { global: { plugins: [pinia] } }).unmount()
@@ -262,6 +260,6 @@ describe('store plugins', () => {
 
     useStore(pinia)
 
-    expect(invokedCount).toBe(1)
+    expect(spy).toHaveBeenCalledTimes(1)
   })
 })
--- a/packages/pinia/__tests__/storePlugins.spec.ts
+++ b/packages/pinia/__tests__/storePlugins.spec.ts
@@ -247,4 +247,21 @@ describe('store plugins', () => {
     store.n++
     expect(spy).toHaveBeenCalledTimes(1)
   })
+
+  it('only executes plugins once after multiple installs', async () => {
+    const pinia = createPinia()
+
+    let invokedCount = 0
+    pinia.use(() => {
+      invokedCount++
+    })
+
+    for (let i = 0; i < 3; i++) {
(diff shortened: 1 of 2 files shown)


packages/pinia/src/createPinia.ts
```diff
@@ -36,6 +36,7 @@ export function createPinia(): Pinia {
           registerPiniaDevtools(app, pinia)
         }
         toBeInstalled.forEach((plugin) => _p.push(plugin))
+        toBeInstalled.splice(0, toBeInstalled.length)
```
posva: let's make it a let and change this to `toBeInstalled = []`

packages/pinia/__tests__/storePlugins.spec.ts
```diff
@@ -247,4 +247,21 @@ describe('store plugins', () => {
...
     expect(spy).toHaveBeenCalledTimes(1)
   })
+
+  it('only executes plugins once after multiple installs', async () => {
+    const pinia = createPinia()
+
+    let invokedCount = 0
+    pinia.use(() => {
+      invokedCount++
+    })
```
posva:
```suggestion
    const spy = jest.fn()
    pinia.use(spy)
```

packages/pinia/__tests__/storePlugins.spec.ts
```diff
@@ -247,4 +247,21 @@ describe('store plugins', () => {
...
+
+    for (let i = 0; i < 3; i++) {
+      mount({ template: 'none' }, { global: { plugins: [pinia] } }).unmount()
+    }
+
+    useStore(pinia)
+
+    expect(invokedCount).toBe(1)
```
posva:
```suggestion
    expect(spy).toHaveBeenCalledTimes(1)
```

posva (changes requested):
Nice! Thanks for test! Let's use a spy to count the amount of invocations.

packages/pinia/src/createPinia.ts
jessevanassen: I wasn't sure if you wanted the original reference to be preserved, so I played it safe.
I've added a new commit with the suggestion.

posva (approved):

PR #754, opened 2021-10-29
docs: fix import `useUserStore`

tohagan:
The [Shared Actions](https://pinia.esm.dev/cookbook/composing-stores.html#shared-actions) example was importing `useCartStore` instead of `useUserStore`.


posva (approved):
Thanks!

PR #764, opened 2021-11-03
fix(nuxt): use `@nuxt/kit`

pi0:
Hi! The latest kit version is now available as `@nuxt/kit`. See [nuxt3-stubs](https://github.com/nuxt/nuxt3-stubs). This will help avoid alias but also reference to non `-edge` version.

I've also replaced `.cjs` build with a proxy since `@nuxt/kit` is an ESM-only package and updated `index` with `module` that will be eventually convention in new module templates.


packages/nuxt/package.json
```diff
@@ -26,16 +26,16 @@
   "sideEffects": false,
   "exports": {
     ".": {
-      "require": "./dist/index.js",
-      "import": "./dist/index.mjs"
+      "import": "./dist/module.mjs",
```
posva, line 29: shouldn't this be renamed to `./dist/index.cjs`? that way siroc will generate the file

packages/nuxt/module.cjs
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const kit = require('@nuxt/kit');
+const vueDemi = require('vue-demi');
+
+const module$1 = kit.defineNuxtModule({
+  name: "pinia",
+  configKey: "pinia",
+  defaults: {
+    disableVuex: true
```
posva: I think this file was included by accident

packages/nuxt/package.json
```diff
@@ -26,16 +26,16 @@
...
+      "require": "./module.cjs"
```
posva, line 30: shouldn't this be renamed to `./dist/index.cjs`? that way siroc will generate the file
```suggestion
      "require": "./dist/index.cjs"
```

packages/nuxt/module.cjs
pi0: Ah I mistakenly replaced it on the last commit.

packages/nuxt/package.json
pi0: This is a manual file and only for legacy nuxt support. Rollup's generated cjs format with require syntax (`require('@nuxt/kit')`) is invalid.

posva: I see, thanks!

posva (approved):
Thanks!

PR #776, opened 2021-11-05
refactor: remove kit dependency

danielroe:
To prevent any issues, what do you think about removing the `@nuxt/kit` dependency for now? This PR converts the module to a non-kit module; we can upgrade later when ready.


posva:
I'm okay removing kit for the moment but this doesn't seem to solve https://github.com/posva/pinia/issues/767 😞 
So I'm not sure this is a good idea.

posva (approved):
After further discussion, it seems to be better to go this way for the moment to better support Nuxt.

Thanks a lot again @danielroe

PR #781, opened 2021-11-07
Only add explicit nuxt alias for nuxt3

### LINKED ISSUE #690: Nuxt.js 2 + nuxt-vite + pinia error
### Reproduction

Thank you so much for such a great state managing lib for Vue.
Pinia works well with Nuxt as well. However, when I tried to use it along with nuxt-vite, an issue happen from the start.
I tried with the fresh install and also, there is a reproduction link to Codesandbox:
[](https://codesandbox.io/s/flamboyant-boyd-z20en)

### Steps to reproduce the behavior

1. Initialize a fresh Nuxt universal app + install the dependencies (the latest versions are used at the moment):
`"dependencies": {
    "@nuxtjs/composition-api": "^0.29.2",
    "@pinia/nuxt": "0.0.2",
    "nuxt": "^2.15.8",
    "pinia": "^2.0.0-rc.9"
  },
  "devDependencies": {
    "nuxt-vite": "^0.2.4"
  }`
2. Add required modules into "buildModules" section of nuxt.config,js:
`buildModules: ["nuxt-vite", "@nuxtjs/composition-api/module", "@pinia/nuxt"],`
3. Run `npm run dev`
4. See the error

### Expected behavior

The app launched correctly with listed modules/configurations.

### Actual behavior

The app won't launch because of the error.
This is comes in the console:
`Uncaught SyntaxError: The requested module '/node_modules/pinia/dist/pinia.cjs?import' does not provide an export named 'defineStore'`

### Additional information

node version is 14.17.3
npm version is 7.24.0

tobiasdiez:
The explicit nuxt alias leads to problems with nuxt/bridge (using vite) as descirbed here https://github.com/posva/pinia/issues/690#issuecomment-956913149.
I'm not completely sure why this alias was added in the first place. Since `pinia` is added to `build.transpile` anyway, the es module build shouldn't be used anyway.

Reverts https://github.com/posva/pinia/commit/438b16dfd3dd5052be437d2e7382f9f4f497eea3
Fixes #690.

### DIFF
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -28,10 +28,6 @@ const module = defineNuxtModule<PiniaNuxtOptions>({
       nuxt.options.features.store = false
     }
 
-    // make sure we use the mjs for pinia so node doesn't complain about using a module js with an extension that is js
-    // but doesn't have the type: module in its packages.json file
-    nuxt.options.alias.pinia = 'pinia/dist/pinia.mjs'
-
     addPlugin({ src: require.resolve('./plugin.mjs') })
 
     // transpile pinia for nuxt 2 and nuxt bridge


posva:
Thanks for the PR, I tested it and it seems not to be enough as it yields

```
Error: [vue-composition-api] must call Vue.use(VueCompositionAPI) before using any function.
```

tobiasdiez:
Maybe it's a good idea to wait until https://github.com/vuejs/pinia/pull/791 is merged? @pi0 what do you think?

tobiasdiez:
With the new version of the nuxt module, everything is working fine and this PR is no longer necessary.

PR #795, opened 2021-11-14
docs: Vuex Migration Docs

### LINKED ISSUE #793: Add a Vuex-to-Pinia Migration Guide to documentation

Looking at this discussion https://github.com/posva/pinia/discussions/787 highlights that a lot of people are converting Vuex stores to Pinia stores. A common thing to do in Vuex is write getters that return state of the same name, of course this is pointless in Pinia. There are a few other gotchas like that when converting from Vuex to Pinia that aren't always obvious.

### Proposed solution

Include a section in the documentation, something like "Vuex Migration Guide", which highlights some of the main steps to do and some of the common gotchas. A rough idea of what a step-by-step guide could look like:

1. Namespacing, usage of id instead, if using other stores with rootState/rootGetters convert those first
2. Convert state, now a function (if it wasn't already)
2. Convert getters, remove useless getters (as in the above discussion), link to type gotchas already documented
3. Convert actions, using `this` instead of context
4. Convert mutations, replace with actions or just assign directly to the store
5. Convert usages of store within components, using helpers or in setup (already documented)

There are some notes at https://pinia.esm.dev/introduction.html#comparison-with-vuex-3-x-4-x and in other places too, but a more step-by-step guide could be beneficial, even if most of it links to already existing documentation.

Another alternative would be to add warnings for these edge cases (for example in the above discussion, when adding getters do a check that it doesn't name collide with any of the state, and warn if it does)

I'd be happy to help in either case, I'm interested in your thoughts on the best approach for handling things like this @posva

BenShelton:
- docs: add vuex migration guide
- docs: add links to vuex migration guide
- docs: fix typo

Close #793

### DIFF
--- a/packages/docs/.vitepress/config.js
+++ b/packages/docs/.vitepress/config.js
@@ -217,7 +217,7 @@ module.exports = {
           link: '/cookbook/',
           children: [
             {
-              text: 'Migration from Vuex',
+              text: 'Migration from Vuex ≤4',
               link: '/cookbook/migration-vuex.html',
             },
             {
--- a/packages/docs/.vitepress/config.js
+++ b/packages/docs/.vitepress/config.js
@@ -236,6 +236,10 @@ module.exports = {
               text: 'Migration from v0/v1 to v2',
               link: '/cookbook/migration-v1-v2.html',
             },
+            {
+              text: 'Migration from Vuex',
+              link: '/cookbook/migration-vuex.html',
+            },
           ],
         },
       ],
--- a/packages/docs/.vitepress/config.js
+++ b/packages/docs/.vitepress/config.js
@@ -217,6 +217,10 @@ module.exports = {
           link: '/cookbook/',
           children: [
             {
+              text: 'Migration from Vuex',
+              link: '/cookbook/migration-vuex.html',
+            },
+            {
               text: 'Hot Module Replacement',
               link: '/cookbook/hot-module-replacement.html',
             },
@@ -235,10 +239,6 @@ module.exports = {
             {
               text: 'Migration from v0/v1 to v2',
               link: '/cookbook/migration-v1-v2.html',
-            },
-            {
-              text: 'Migration from Vuex',
-              link: '/cookbook/migration-vuex.html',
             },
           ],
         },
--- a/packages/docs/cookbook/index.md
+++ b/packages/docs/cookbook/index.md
@@ -1,8 +1,8 @@
 # Cookbook
 
+- [Migrating from Vuex](./migration-vuex.md): A migration guide for converting Vuex ≤4 projects.
 - [HMR](./hot-module-replacement.md): How to activate hot module replacement and improve the developer experience.
 - [Testing Stores (WIP)](./testing.md): How to unit test Stores and mock them in component unit tests.
 - [Composing Stores](./composing-stores.md): How to cross use multiple stores. e.g. using the user store in the cart store.
 - [Options API](./options-api.md): How to use Pinia without the composition API, outside of `setup()`.
 - [Migrating from 0.0.7](./migration-0-0-7.md): A migration guide with more examples than the changelog.
-- [Migrating from Vuex](./migration-vuex.md): A migration guide for converting Vuex ≤4 projects.
--- a/packages/docs/cookbook/index.md
+++ b/packages/docs/cookbook/index.md
@@ -1,6 +1,6 @@
 # Cookbook
 
-- [Migrating from Vuex](./migration-vuex.md): A migration guide for converting Vuex ≤4 projects.
+- [Migrating from Vuex ≤4](./migration-vuex.md): A migration guide for converting Vuex ≤4 projects.
 - [HMR](./hot-module-replacement.md): How to activate hot module replacement and improve the developer experience.
 - [Testing Stores (WIP)](./testing.md): How to unit test Stores and mock them in component unit tests.
 - [Composing Stores](./composing-stores.md): How to cross use multiple stores. e.g. using the user store in the cart store.
--- a/packages/docs/cookbook/index.md
+++ b/packages/docs/cookbook/index.md
@@ -5,3 +5,4 @@
 - [Composing Stores](./composing-stores.md): How to cross use multiple stores. e.g. using the user store in the cart store.
 - [Options API](./options-api.md): How to use Pinia without the composition API, outside of `setup()`.
 - [Migrating from 0.0.7](./migration-0-0-7.md): A migration guide with more examples than the changelog.
+- [Migrating from Vuex](./migration-vuex.md): A migration guide for converting Vuex ≤4 projects.
--- a/packages/docs/cookbook/migration-vuex.md
+++ b/packages/docs/cookbook/migration-vuex.md
@@ -44,7 +44,7 @@ For large projects you may wish to do this conversion module by module rather th
 
 ## Converting a Single Module
 
-Here is a complete example of the before and after of converting a Vuex module to a Pinia store, see below for a step-by-step guide:
+Here is a complete example of the before and after of converting a Vuex module to a Pinia store, see below for a step-by-step guide. The Pinia example uses an option store as the structure is most similar to Vuex:
 
 ```ts
 // Vuex module in the 'auth/user' namespace
@@ -68,20 +68,22 @@ const storeModule: Module<State, RootState> = {
   getters: {
     firstName: (state) => state.firstName,
     fullName: (state) => `${state.firstName} ${state.lastName}`,
-    loggedIn: (state) => !!state.userId,
+    loggedIn: (state) => state.userId !== null,
     // combine with some state from other modules
     fullUserDetails: (state, getters, rootState, rootGetters) => {
       return {
         ...state,
         fullName: getters.fullName,
+        // read the state from another module named `auth`
         ...rootState.auth.preferences,
+        // read a getter from a namespaced module called `email` nested under `auth`
         ...rootGetters['auth/email'].details
       }
     }
   },
   actions: {
     async loadUser ({ state, commit }, id: number) {
-      if (state.userId) throw new Error('Already logged in')
+      if (state.userId !== null) throw new Error('Already logged in')
       const res = await api.user.load(id)
       commit('updateUser', res)
     }
@@ -126,9 +128,9 @@ export const useAuthUserStore = defineStore('auth/user', {
   getters: {
     // firstName getter removed, no longer needed
     fullName: (state) => `${state.firstName} ${state.lastName}`,
-    loggedIn: (state) => !!state.userId,
+    loggedIn: (state) => state.userId !== null,
     // must define return type because of using `this`
-    fullUserDetails: (state): FullUserDetails => {
+    fullUserDetails (state): FullUserDetails {
       // import from other stores
       const authPreferencesStore = useAuthPreferencesStore()
       const authEmailStore = useAuthEmailStore()
@@ -152,7 +154,7 @@ export const useAuthUserStore = defineStore('auth/user', {
   actions: {
     // no context as first argument, use `this` instead
     async loadUser (id: number) {
-      if (this.userId) throw new Error('Already logged in')
+      if (this.userId !== null) throw new Error('Already logged in')
       const res = await api.user.load(id)
       this.updateUser(res)
     },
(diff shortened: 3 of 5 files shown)


packages/docs/.vitepress/config.js
```diff
@@ -236,6 +236,10 @@ module.exports = {
               text: 'Migration from v0/v1 to v2',
               link: '/cookbook/migration-v1-v2.html',
             },
+            {
```
posva: Let's put it first in the children array

packages/docs/cookbook/index.md
```diff
@@ -5,3 +5,4 @@
 - [Composing Stores](./composing-stores.md): How to cross use multiple stores. e.g. using the user store in the cart store.
 - [Options API](./options-api.md): How to use Pinia without the composition API, outside of `setup()`.
 - [Migrating from 0.0.7](./migration-0-0-7.md): A migration guide with more examples than the changelog.
+- [Migrating from Vuex](./migration-vuex.md): A migration guide for converting Vuex ≤4 projects.
```
posva: let's move this up too

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
+# Migrating from Vuex
```
posva: I'm thinking we should maybe specify Vuex 4 even though Vuex 5 is not out yet

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Preparation
+
+First, follow the [Getting Started guide](../getting-started.md) to install Pinia.
+
+## Restructuring Modules to Stores
+
+Vuex has the concept of a single store with multiple _modules_. These modules can optionally be namespaced and even nested within each other.
+
+The easiest way to transition that concept to be used with Pinia is that each module you used previously is now a _store_. Each store has an `id` which is similar to a namespace. Nested modules can also each become their own store. Stores that depend on each other will simply import the other store.
```
posva: You could also say that each store is namespaced by design and thus requires a unique identifier.

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+
+How you choose to restructure your Vuex modules into Pinia stores is entirely up to you, but here is one suggestion:
+
+```sh
```
posva: This should probably be just
````suggestion
```
````

File: packages/docs/cookbook/migration-vuex.md [hunk 6]
```diff
@@ -0,0 +1,271 @@
...
+# Vuex example (assuming namespaced modules)
+src/
+├─ store/
+│  ├─ index.js          # Initializes Vuex, imports modules
+│  ├─ modules/
+│  │  ├─ module1.js     # 'module1' namespace
+│  │  ├─ nested/
+│  │  │  ├─ index.js    # 'nested' namespace, imports module2 & module3
+│  │  │  ├─ module2.js  # 'nested/module2' namespace
+│  │  │  ├─ module3.js  # 'nested/module3' namespace
```
posva: I think we should use a command like `tree` to properly display the structure. I can handle it at the end if you want

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+
+# Pinia equivalent, note ids match previous namespaces
+src/
+├─ stores/
+│  ├─ index.js          # Initializes Pinia, does not import stores
```
posva: We could also say this is optional:
```suggestion
│  ├─ index.js          # (Optional) Initializes Pinia, does not import stores
```

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+│  ├─ module1.js        # 'module1' id
+│  ├─ nested.js         # 'nested' id
+│  ├─ nested-module2.js # 'nested/module2' id
+│  ├─ nested-module3.js # 'nested/module3' id
+```
+
+This creates a flat structure for stores but also preserves the previous namespacing with equivalent `id`s. If you had some state/getters/actions/mutations in the root of the store (in the `store/index.js` file of Vuex) you may wish to create another store called something like `root` which holds all that information.
+
+For large projects you may wish to do this module by module rather than converting everything at once. You can actually mix Pinia and Vuex together during the migration so this approach can also work.
```
posva: I think it's worth noting somewhere that the folder is now named `stores`, using plural, as it's easy to overlook by people and it helps having both Vuex and Pinia while migrating.

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+  getters: {
+    firstName: (state) => state.firstName,
+    fullName: (state) => `${state.firstName} ${state.lastName}`,
+    loggedIn: (state) => !!state.userId,
+    // combine with some state from other modules
+    fullUserDetails: (state, getters, rootState, rootGetters) => {
+      return {
+        ...state,
+        fullName: getters.fullName,
+        ...rootState.auth.preferences,
```
posva, line 78: I think we can make this more explicit:
```suggestion
        // read the state from another module named `auth`
        ...rootState.auth.preferences,
```

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+        ...rootGetters['auth/email'].details
```
posva, line 80: And here we could have something similar

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+  namespaced: true,
+  state: {
+    firstName: '',
+    lastName: '',
+    userId: null
+  },
+  getters: {
+    firstName: (state) => state.firstName,
+    fullName: (state) => `${state.firstName} ${state.lastName}`,
+    loggedIn: (state) => !!state.userId,
```
posva: should probably be
```suggestion
    loggedIn: (state) => state.userId != null,
```

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+  // convert to a function
+  state: (): State => ({
+    firstName: '',
+    lastName: '',
+    userId: null
+  }),
+  getters: {
+    // firstName getter removed, no longer needed
+    fullName: (state) => `${state.firstName} ${state.lastName}`,
+    loggedIn: (state) => !!state.userId,
```
posva:
```suggestion
    loggedIn: (state) => state.userId != null,
```

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+    // must define return type because of using `this`
+    fullUserDetails: (state): FullUserDetails => {
```
posva: this cannot be an arrow function if it uses `this`:

```suggestion
    fullUserDetails(state): FullUserDetails {
```

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+      //   fullName: this.fullName,
+      //   ...vuexStore.state.auth.preferences,
+      //   ...vuexStore.getters['auth/email'].details
+      // }
+    }
+  },
+  actions: {
+    // no context as first argument, use `this` instead
+    async loadUser (id: number) {
+      if (this.userId) throw new Error('Already logged in')
```
posva:
```suggestion
      if (this.userId != null) throw new Error('Already logged in')
```

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+})
+```
+
+Let's break the above down into steps:
+
+1. Add a required `id` for the store, you may wish to keep this the same as the namespace before
+2. Convert `state` to a function if it was not one already
+3. Convert `getters`
+    1. Remove any getters that return state under the same name (eg. `firstName: (state) => state.firstName`), these are not necessary as you can access any state directly from the store instance
+    2. If you need to access other getters, they are on `this` instead of using the second argument. Note you will need to specify a return type if you do this because of TS limitations, see [here](../core-concepts/getters.md#accessing-other-getters) for more details
```
posva: Let's add something along the lines _and you will have to use a regular function instead of an arrow function_

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+    3. If using `rootState` or `rootGetters` arguments, replace them by importing the other store directly, or if they still exist in Vuex then access them directly from Vuex
+4. Convert `actions`
+    1. Remove the first `context` argument from each action. Everything should be accessible from `this` instead
+    2. If using other stores either import them directly or access them on Vuex, the same as for getters
+5. Convert `mutations`
+    1. Mutations do not exist any more. These can be converted to `actions` instead, or you can just assign directly to the store within your components (eg. `userStore.firstName = 'First'`)
+    2. If converting to actions, remove the first `state` argument and replace any assignments with `this` instead
+    3. A common mutation is to reset the state back to its initial state. This is built in functionality with the store's `$reset` method
```
posva: I wonder if it should be pointed out somehow that the reset functionality only exists for option store (which isn't mentioned and I think that's normal)

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+// Pinia
+import { defineComponent, computed } from 'vue'
+import { useAuthUserStore } from '@/stores/auth-user'
+
+export default defineComponent({
+  setup () {
+    const authUserStore = useAuthUserStore()
+
+    const firstName = computed(() => authUserStore.firstName)
+    const fullName = computed(() => authUserStore.fullName)
```
posva, line 231: I think it's worth noting you can just return the `store` inside of setup and directly use `store.firstName`. I also think we should add the `return` statement to both examples.

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+  }
+})
+```
+
+## Usage Outside Components
+
+Updating usage outside of components should be simple as long as you're careful to _not use a store outside of functions_. Here is an example of using the store in a `vuex-router` hook:
```
posva: Isn't it:
```suggestion
Updating usage outside of components should be simple as long as you're careful to _not use a store outside of functions_. Here is an example of using the store in a Vue Router navigation guard:
```

packages/docs/cookbook/migration-vuex.md
```diff
@@ -0,0 +1,271 @@
...
+
+More details can be found [here](../core-concepts/outside-component-usage.md).
+
+## Advanced Vuex Usage
+
+In the case your Vuex store using some of the more advanced features it offers, here is some guidance on how to accomplish the same in Pinia. Some of these points are already covered in [this comparison summary](../introduction.md#comparison-with-vuex-3-x-4-x).
+
+### Dynamic Modules
+
+There is no need to dynamically register modules in Pinia. Stores are dynamic by default and are only registered when they are needed. If a store is never used, it will never be "registered".
```
posva:
```suggestion
There is no need to dynamically register modules in Pinia. Stores are dynamic by design and are only registered when they are needed. If a store is never used, it will never be "registered".
```

posva (changes requested):
Great job!

packages/docs/cookbook/migration-vuex.md
BenShelton: Reworded the second paragraph slightly to include that.

packages/docs/cookbook/migration-vuex.md
BenShelton: I used this so that the comments would appear dimmer, which works in my IDE but not on the actual site. Changing this to `bash` does though, see what you think. With no language it all appears white which makes it a little harder to read IMO.

packages/docs/cookbook/migration-vuex.md
BenShelton: This was generated using https://ascii-tree-generator.com/

Replaced with `tree` output

packages/docs/cookbook/migration-vuex.md
BenShelton: Added a note here and also at the start of this section stating that we're using option store as it's the most similar to Vuex.

packages/docs/cookbook/migration-vuex.md
BenShelton: Yeh good catch, dunno where that came from, my mind had stopped working at this point, writing docs is hard 😆

posva (approved):
Nice! Thanks!

packages/docs/cookbook/index.md
```diff
@@ -1,5 +1,6 @@
 # Cookbook
 
+- [Migrating from Vuex](./migration-vuex.md): A migration guide for converting Vuex ≤4 projects.
```
posva:
```suggestion
- [Migrating from Vuex ≤4](./migration-vuex.md): A migration guide for converting Vuex ≤4 projects.
```

packages/docs/.vitepress/config.js
```diff
@@ -216,6 +216,10 @@ module.exports = {
           text: 'Cookbook',
           link: '/cookbook/',
           children: [
+            {
+              text: 'Migration from Vuex',
```
posva:
```suggestion
              text: 'Migration from Vuex ≤4',
```

posva (commented):
I'm just adding ≤4 to the titles as well

PR #811, opened 2021-11-19
fix(testing): typo when detecting jest existence

bodograumann:
Currently `createTestingPinia` is only usable in jest context.


posva (approved):
Thanks!

PR #836, opened 2021-11-25
docs: fix typo in actions.md

shmarts:
small update to the actions docs - the action further down refers to a state property which doesn't exist


posva (approved):
Thanks

PR #839, opened 2021-11-26
docs: update readme

l4dybird:
docs update increment methods


posva:
Thanks but you don't need an increment action with pinia!

PR #846, opened 2021-11-28
fix: shouldHydrate if not in skipHydrateMap

### LINKED ISSUE #845: Nuxt 2/Bridge: state is not hydrated when using setupStore
### Reproduction

This only appears to be an issue when using Vue 2 as the `shouldHydrate` method is working properly with Vue 3.

1. Create a store using a setup function
2. The store is hydrated on the server
3. The store is empty on the client

### Expected behavior

The store should be hydrated on the client using Pinia's initial state.

### Actual behavior

Pinia's initial state is set correctly on the server but the hydration is skipped on the client because there seems to be a typo in the `shouldHydrate` function when using Vue 2: I believe the expected behavior of that function is to check whether a given object is NOT in `skipHydrateMap` and it is currently checking if it IS in `skipHydrateMap`.

```js
function shouldHydrate(obj) {
    return isVue2
        ? skipHydrateMap.has(obj)
        : !isPlainObject(obj) || !obj.hasOwnProperty(skipHydrateSymbol);
}
```
should probably be:
```js
function shouldHydrate(obj) {
    return isVue2
        ? !skipHydrateMap.has(obj)
        : !isPlainObject(obj) || !obj.hasOwnProperty(skipHydrateSymbol);
}
```

Unless I'm mistaken? In any case, thanks for your amazing work 🙂

### Additional information

I'm providing a PR referencing that issue.

yassilah:
fixes #845

### DIFF
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -88,7 +88,7 @@ export function skipHydrate<T = any>(obj: T): T {
 
 function shouldHydrate(obj: any) {
   return isVue2
-    ? /* istanbul ignore next */ skipHydrateMap.has(obj)
+    ? /* istanbul ignore next */ !skipHydrateMap.has(obj)
     : !isPlainObject(obj) || !obj.hasOwnProperty(skipHydrateSymbol)
 }
 


posva (approved):
Thanks! Nice catch

PR #878, opened 2021-12-06
feat: expose providePinia to use multiple root stores

### LINKED ISSUE #870: Allow changing pinia root store for component scope

When documenting a component with storybook and its docs addon, multiple stories are rendered in one MDX document. That means they are technically part of the same Vue application.
Semantically however, each story represents a very specific state of the component. Usually these components would not access any global state (and thus pinia) at all, but sometimes it is necessary. That means each of the story components needs its own, independant “global” state.

### Proposed solution

Pinia should allow to change the root store Pinia instance which is injected into a component and its descendents.
Technically this could be done simply by exposing the `piniaSymbol` and than I could use `provide(piniaSymbol, createPinia())`.

On the other hand it might be more prudent to keep the symbol private and expose a convenience function `providePinia(pinia) { provide(piniaSymbol, pinia ?? createPinia()) }`.

It is also possible to patch alias fields into the storybook webpack config and mock the store definition module which provides the `useMyStore` function for each store individually. Then each story could set a currentMock property in those modules.
This would be much more complex however, harder to understand and possibly fragile.

bodograumann:
Implements #870.

### DIFF
--- a/packages/pinia/__tests__/multipleRoots.spec.ts
+++ b/packages/pinia/__tests__/multipleRoots.spec.ts
@@ -1,0 +1,98 @@
+import { createPinia, defineStore, providePinia } from '../src'
+import { mount } from '@vue/test-utils'
+import { defineComponent } from 'vue'
+
+describe('Multiple Roots', () => {
+  function defineMyStore() {
+    return defineStore({
+      id: 'main',
+      state: () => ({
+        n: 0,
+      }),
+    })
+  }
+
+  it('uses the same root in child components by default', () => {
+    expect.assertions(2)
+    const pinia = createPinia()
+    const useStore = defineMyStore()
+
+    const ChildComponent = defineComponent({
+      template: 'no',
+      setup() {
+        const store = useStore()
+        expect(store.n).toBe(1)
+      },
+    })
+
+    mount(
+      {
+        template: '<ChildComponent />',
+        components: { ChildComponent },
+        setup() {
+          const store = useStore()
+          expect(store.n).toBe(0)
+          store.n++
+        },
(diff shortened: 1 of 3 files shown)


posva:
Thanks! I will have to give this more thought as there might be some edge cases with `setActivePinia()` and how it is used across stores

bodograumann:
Is there some way I can help with that?
If you have specific scenarios in mind, I could add tests for them.

posva:
Currently, there are still a few issues:

- It shouldn't change the activePinia
- It cannot use plugins (they require the `pinia._a` property but even if it was set, it would still mean plugins could only be added after calling providePlugin

Overall, I think exposing the symbol and letting the user use at their own risk might be a better idea. Otherwise, a proper solution would be a custom pinia similar to the testing one

posva:
@bodograumann would exposing the symbol be enough? 

I added a couple of tests to explain what I meant. Maybe `createPinia()` could detect it's a nested pinia but that would also mean it would add to the library size so it should be really minimal.

There is also a concern about devtools. Right now, they expect **no more than one** pinia to be created for an app.

bodograumann:
I have a hard time wrapping my head around the whole context right now.
In the [feature request](https://github.com/vuejs/pinia/issues/870) some more details are layed out however.
My past self thought that exposing the symbol is good enough.
Also, I’m only using this functionality in the context of storybook+createTestingPinia.
If it helps, I could create an example project.

posva:
I know it's been a very long time, but since this hasn't advanced, maybe exposing the symbol at the user's own risk is just the simpler version for the moment (#2973 )

PR #893, opened 2021-12-12
fix(onAction): more onAction after can run

MrChenYZ:
The main problem is that the parent and child components $onAction at the same time will only trigger the last after callback, but in the actual scenario, we may need to use this after callback.

### DIFF
--- a/packages/pinia/__tests__/onAction.spec.ts
+++ b/packages/pinia/__tests__/onAction.spec.ts
@@ -10,8 +10,12 @@ describe('Subscriptions', () => {
       id: 'main',
       state: () => ({
         user: 'Eduardo',
+        theme: 'black',
       }),
       actions: {
+        changeTheme(theme: string) {
+          this.theme = theme
+        },
         direct(name: string) {
           this.user = name
         },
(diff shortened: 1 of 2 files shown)


MrChenYZ:
Multiple $onAction are registered, but only the last after is running.

posva:
You’re supposed to call `after` only once. Is that what you meant to fix?
If not I think a boiled down repro would help because this dix complexifies a lot the implémentation

posva:
You’re supposed to call `after` only once. Is that what you meant to fix?
If not I think a boiled down repro would help because this dix complexifies a lot the implémentation

MrChenYZ:
> You’re supposed to call `after` only once. Is that what you meant to fix? If not I think a boiled down repro would help because this dix complexifies a lot the implémentation

I use `onAction` in multiple places, why only one `after` is executed? So if you follow the current thinking, why do you need to execute `onAction` multiple times instead of executing multiple times his callback? Because I may have done monitoring in different components, I may need to do some callback processing. For example, I use a `pop-up window` to log in, and I want to monitor user changes in the `header component`, and then make some other data loading requests. Is this unreasonable? At the same time, I also provided 3 scenarios for testing. I think scenario 3 should be used frequently in actual development.
code

posva:
I'm revisiting the tests and in this case

```ts
    const spy = jest.fn()
    store.$onAction(({ after }) => {
      after(spy)
    })
    store.$onAction(({ after }) => {
      after(spy)
    })
```

the test is not valid because you should not be passing the same function twice to the `after` of one given store. Do you have a use case for that? We could maybe warn against this though.

However, this one should work and currently it isn't:

```ts
    const spy1 = jest.fn()
    const spy2 = jest.fn()
    store.$onAction(({ after }) => {
      after(spy1)
    })
    store.$onAction(({ after }) => {
      after(spy2)
    })

    await expect(store.asyncUpperName()).resolves.toBe('EDUARDO')
    expect(spy2).toHaveBeenCalledTimes(1)
    expect(spy1).toHaveBeenCalledTimes(1)
```

MrChenYZ:
Currently, the plugin allows multiple `onAction` subscriptions to be executed, and as a result, only one `after` callback is triggered. But there is a scenario where the `parent component` triggers the action, the `child component` gets the return value through after, and the `parent component` cannot get the action return value. For `child components`, this after return value is `useless`, but for `parent components`, it is very `useful`. The child component may use this `after` callback to do some other things, and its function can be completely isolated from the parent component. Therefore, I think this is an unavoidable user behavior. If the user can write after, it means it has his purpose. We should not prohibit him.

posva:
Thank you for this, unfortunately, I had to revisit quite a few things, please refer to the commit to see the differences

PR #900, opened 2021-12-14
test(storePlugins): remove no use type S

so11y:
remove no use generics arg


posva:
Thanks but this is correct

PR #909, opened 2021-12-18
chore(docs): added pnpm installation guide

dammy001:


posva:
Thanks but I prefer keeping just two options. People using pnpm know how to use it

PR #914, opened 2021-12-21
docs: use new docs URL

yyx990803:
I've set up a new deployment using our vuejs team account with a new domain at https://pinia.vuejs.org. 🎉

You should already have access to the deployment!

### DIFF
--- a/packages/pinia/src/devtools/plugin.ts
+++ b/packages/pinia/src/devtools/plugin.ts
@@ -56,9 +56,9 @@ export function registerPiniaDevtools(app: DevtoolsApp, pinia: Pinia) {
     {
       id: 'dev.esm.pinia',
       label: 'Pinia 🍍',
-      logo: 'https://pinia.esm.dev/logo.svg',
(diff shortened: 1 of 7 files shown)


packages/pinia/src/devtools/plugin.ts
```diff
@@ -56,9 +56,9 @@ export function registerPiniaDevtools(app: DevtoolsApp, pinia: Pinia) {
     {
       id: 'dev.esm.pinia',
```
posva, line 57:
```suggestion
      id: 'org.vuejs.pinia',
```

packages/pinia/src/devtools/plugin.ts
```diff
@@ -265,9 +265,9 @@ function addStoreToDevtools(app: DevtoolsApp, store: StoreGeneric) {
...
```
posva, line 266:
```suggestion
      id: 'org.vuejs.pinia',
```

posva (approved):
Thanks! Got the redirect working too

PR #915, opened 2021-12-21
refactor(types): set gereric for state

heart-er:
we should set generic for state so that state wont loose type

In Vue, effect run's type has a generic input and return the same generic
so, if not set genric, state's type will like:
image

so if we set generic, state's type will not loose:
image


posva:
What is this actually fixing? Can you provide a failing test or reproduction?
The type shouldn't matter here as the one in `Pinia` is the one kept anyway

heart-er:
oh, sorry, this is like a type optimization, because in VSCode, type is displayed as a loose type, so I see it as an internal type definition problem.

heart-er:
Although TypeScript compares Type's shape, but we should avoid loose types as much as possible.

PR #924, opened 2021-12-24
feat(types): support IDE features for store context

johnsoncodehk:
Fixed:

```ts
const useFoo = defineStore('foo', () => ({ bar: 123 }));
const foo = useFoo();
foo.bar // <-- renaming, goto definition, find references not working for bar
```

### DIFF
--- a/packages/pinia/src/types.ts
+++ b/packages/pinia/src/types.ts
@@ -546,31 +546,43 @@ export type _ActionsTree = Record<string, _Method>
 /**
  * @internal
  */
+declare type _ExtractStateFromSetupStore_Keys<SS> = keyof { [K in keyof SS as SS[K] extends _Method | ComputedRef ? never : K]: any }
+
+/**
+ * @internal
+ */
+declare type _ExtractActionsFromSetupStore_Keys<SS> = keyof { [K in keyof SS as SS[K] extends _Method ? K : never]: any }
+
+/**
+ * @internal
+ */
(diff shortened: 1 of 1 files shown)


packages/pinia/src/types.ts
```diff
@@ -543,34 +543,46 @@ export type _GettersTree<S extends StateTree> = Record<
...
+
+/**
+ * @internal
+ */
+declare type _ExtractActionsFromSetupStore_Keys<SS> = keyof { [K in keyof SS as SS[K] extends _Method ? K : never]: any }
+
+/**
+ * @internal
+ */
+declare type _ExtractGettersFromSetupStore_Keys<SS> = keyof { [K in keyof SS as SS[K] extends ComputedRef ? K : never]: any }
```
posva: shouldn't this and others be a regular types?

posva (changes requested):
Thanks a lot! Can you make sure the tests pass? Formatting seems to be failing

packages/pinia/src/types.ts
johnsoncodehk: Don't know what I can't make Prettier working in this project, I just copy the code to other project and do format. 😅

> shouldn't this and others be a regular types?

I think these just attached types of `_ExtractStateFromSetupStore`, `_ExtractActionsFromSetupStore`, `_ExtractGettersFromSetupStore` but not regular types.

PR #953, opened 2022-01-12
docs: vars references

juliolmuller:
Improving documentation writing.

### DIFF
--- a/packages/docs/cookbook/composing-stores.md
+++ b/packages/docs/cookbook/composing-stores.md
@@ -5,37 +5,37 @@ Composing stores is about having stores that use each other and there is one rul
 If **two or more stores use each other**, they cannot create an infinite loop through _getters_ or _actions_. They cannot **both** directly read each other state in their setup function:
 
 ```js
-const useA = defineStore('a', () => {
-  const b = useB()
+const useX = defineStore('x', () => {
+  const y = useY()
 
-  // ❌ this is not possible because 'b' also tries to read 'a.name'
-  b.name
+  // ❌ This is not possible because y also tries to read x.name
+  y.name
 
   function doSomething() {
-    // ✅ Read 'b' properties in computed or actions
-    const bName = b.name
+    // ✅ Read y properties in computed or actions
+    const yName = y.name
     // ...
   }
 
   return {
-    name: ref('I am A'),
+    name: ref('I am X'),
   }
 })
 
-const useB = defineStore('b', () => {
-  const a = useA()
+const useY = defineStore('y', () => {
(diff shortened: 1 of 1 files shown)


posva (changes requested):
I prefer the other version without quotes. But if you want, you can find a more interesting (yet simple) example where both stores can use each other so we can use more explicit names. It doesn't have to be realistic, it can be playing with words, pop culture, etc. e.g. using codependent things like maybe Romeo and Juliet

soc221b:
I have this example, but maybe the original is the best because it's so simple.

```ts
import { ref } from 'vue'
import { defineStore } from 'pinia'

export const useUserStore = defineStore('users', () => {
  const users = ref([
    {
      id: 1,
      name: 'Leanne Graham',
      postIds: [1, 2],
    },
  ])

  const posts = usePostStore()

  // ❌ this is not possible because it directly read `posts`' states or actions in setup function
  // console.log(posts.getPostTitle(1))

  function getUserName(userId: number) {
    return users.value.find((user) => user.id === userId)?.name
  }

  function getPostTitlesByUserId(userId: number) {
    // ✅ Read posts properties in computed or actions
    const user = users.value.find((user) => user.id === userId)
    if (user === undefined) return undefined

    return user.postIds.map((postId) => posts.getPostTitle(postId))
  }

  return {
    users,
    getUserName,
    getPostTitlesByUserId,
  }
})

export const usePostStore = defineStore('posts', () => {
  const posts = ref([
    {
      userId: 1,
      id: 1,
      title:
        'sunt aut facere repellat provident occaecati excepturi optio reprehenderit',
    },
    {
      userId: 1,
      id: 2,
      title: 'qui est esse',
    },
  ])

  const users = useUserStore()

  // ❌ this is not possible because it directly read `users`' states or actions in setup function
  // console.log(users.getUserName(1))

  function getPostTitle(postId: number) {
    return posts.value.find((post) => post.id === postId)?.title
  }

  function getUserNameByPostId(postId: number) {
    // ✅ Read users properties in computed or actions
    const post = posts.value.find((post) => post.id === postId)
    if (post === undefined) return undefined

    return users.getUserName(post.userId)
  }

  return {
    posts,
    getPostTitle,
    getUserNameByPostId,
  }
})
```

juliolmuller:
I agree, @iendeavor ! The original example is very simple and straight-forward, I think we should stick with that. A more realistic, but complementary, example would be helpful though, I would say.

What about changing the variable to `x` and `y`, @posva , so we get rid of the quotes and avoid confusion with the article *a*?

PR #960, opened 2022-01-14
docs: Moved 20px margin from child to the parent.

assasin0076:
image

its a bug right there that i fixed ^


posva:
What page did you see this on? The menu should not be empty

assasin0076:
> What page did you see this on? The menu should not be empty

https://pinia.vuejs.org/api/interfaces/pinia.mapstorescustomization.html#interface-mapstorescustomization Its right here

posva:
Thank you but this creates too many pages on other pages which are more visited. Since we are moving to vue theme very soon, I don't think it's worth changing this as it does require a few changes to other paddings as well

PR #967, opened 2022-01-17
fix: check HTMLAnchorElement in saveAs for mini-program (#966)

### LINKED ISSUE #966: HTMLAnchorElement causes the mini program to crash
截屏2022-01-18 上午3 12 34

Like #547 

There is no dom environment in the mini program, which is used by hundreds of millions of people in China, and worse, many companies in China (e.g. Baidu, Byte, Alipay) are making their own mini program. They may be bad, but they do run on js, and there are companies making their own systems using js as an application development language (Huawei for example).
Correspondingly, there are also companies that are engaging in vue/react into various frameworks for small programs (WeChat team is also doing it)

Vue3 already works well in Taro3, I've been using patch-package to deal with this problem before (sorry to mention it now), but after recently changing to pnpm, patch-package has a lot of problems.

Avoiding the Dom Api as much as possible might make compatibility easier to do?

odex21:
fix #966

### DIFF
--- a/packages/pinia/src/devtools/file-saver.ts
+++ b/packages/pinia/src/devtools/file-saver.ts
@@ -112,8 +112,10 @@ export type SaveAs =
 
 export const saveAs: SaveAs = !IS_CLIENT
   ? () => {} // noop
-  : // Use download attribute first if possible (#193 Lumia mobile) unless this is a macOS WebView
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks! Note this file-saver code is removed in production as it's only used during dev

PR #976, opened 2022-01-20
feat(testing): add pre-configuration for vitest

tkint:
Just adding a pre-configuration for vitest in `createTestingPinia`

### DIFF
--- a/packages/testing/src/testing.ts
+++ b/packages/testing/src/testing.ts
@@ -46,7 +46,7 @@ export interface TestingOptions {
 
   /**
    * Function used to create a spy for actions and `$patch()`. Pre-configured
-   * with `jest.fn()` in jest projects or `vitest.fn()` in vitest projects
+   * with `jest.fn()` in jest projects or `vitest.fn()` in vitest projects.
    */
   createSpy?: (fn?: (...args: any[]) => any) => (...args: any[]) => any
 }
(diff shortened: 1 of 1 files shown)


packages/testing/src/testing.ts
```diff
@@ -46,7 +46,7 @@ export interface TestingOptions {
 
   /**
    * Function used to create a spy for actions and `$patch()`. Pre-configured
-   * with `jest.fn()` in jest projects.
+   * with `jest.fn()` in jest projects or `vitest.fn()` in vitest projects
```
posva:
```suggestion
   * with `jest.fn()` in jest projects or `vitest.fn()` in vitest projects.
```

packages/testing/src/testing.ts
```diff
@@ -64,8 +64,8 @@ export interface TestingPinia extends Pinia {
  * Creates a pinia instance designed for unit tests that **requires mocking**
  * the stores. By default, **all actions are mocked** and therefore not
  * executed. This allows you to unit test your store and components separately.
- * You can change this with the `stubActions` option. If you are using jest,
- * they are replaced with `jest.fn()`, otherwise, you must provide your own
+ * You can change this with the `stubActions` option. If you are using jest or vitest,
+ * they are replaced with basic `fn()`, otherwise, you must provide your own
```
posva: let's revert this changes, having one example with jest is enough here

packages/testing/src/testing.ts
```diff
@@ -89,7 +89,7 @@ export function createTestingPinia({
 
   plugins.forEach((plugin) => pinia.use(plugin))
 
-  const createSpy = _createSpy || (typeof jest !== 'undefined' && jest.fn)
+  const createSpy = _createSpy || (typeof jest !== 'undefined' && jest.fn) || (typeof vitest !== 'undefined' && vitest.fn)
```
posva, line 92: Isn't it 
```suggestion
  const createSpy = _createSpy || (typeof jest !== 'undefined' && jest.fn) || (typeof vi !== 'undefined' && vi.fn)
```

posva (changes requested):

packages/testing/src/testing.ts
tkint: Both of them are valid: https://github.com/vitest-dev/vitest/blob/main/packages/vitest/globals.d.ts

posva (changes requested):
Can you fix the tests? (formatting so far)

posva:
Thanks, I merged this locally 44f0762

PR #990, opened 2022-01-25
fix(subscriptions): allow removing subscriptions inside them

LemonAppleMo:
I'm sorry, my English is not very good. For a brief expression, I only used $subscribe once at that time, which will lead to exceptions in the subsequent $subscribe. I try to fix it in this way.

### DIFF
--- a/packages/pinia/src/subscriptions.ts
+++ b/packages/pinia/src/subscriptions.ts
@@ -30,7 +30,7 @@ export function triggerSubscriptions<T extends _Method>(
   subscriptions: T[],
   ...args: Parameters<T>
 ) {
-  ;[...subscriptions].forEach((callback) => {
+  subscriptions.slice().forEach((callback) => {
     callback(...args)
   })
 }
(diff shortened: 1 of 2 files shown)


posva:
This doesn't seem to be necessary. Do you have a boiled down repro?

LemonAppleMo:
Example: https://github.com/LemonAppleMo/pinia-subscribe

When I subscribe to the store in two components at the same time, clearing one affects the other.

packages/pinia/src/subscriptions.ts
```diff
@@ -30,7 +30,7 @@ export function triggerSubscriptions<T extends _Method>(
   subscriptions: T[],
   ...args: Parameters<T>
 ) {
-  subscriptions.forEach((callback) => {
+  ;[...subscriptions].forEach((callback) => {
```
posva:
```suggestion
  subscriptions.slice().forEach((callback) => {
```

PR #1011, opened 2022-01-30
docs: description to use composition API approach

davidmeirlevy:
added description to use composition API approach for Pinia's store definition.

### DIFF
--- a/packages/docs/core-concepts/index.md
+++ b/packages/docs/core-concepts/index.md
@@ -14,6 +14,21 @@ export const useStore = defineStore('main', {
 
 This _name_, also referred as _id_, is necessary and is used by Pinia to connect the store to the devtools. Naming the returned function _use..._ is a convention across composables to make its usage idiomatic.
 
+Note that you can also define a store using a setup function instead, similar to Vue's Composition API:
+
+```js
+import { defineStore } from 'pinia'
+
+// useStore could be anything like useUser, useCart
+// the first argument is a unique id of the store across your application
+export const useStore = defineStore('main', () => {
+  return {
+    // any functions, computed, ref and reactive objects as you like.
+  }
+})
+```
+
+
 ## Using the store
 
 We are _defining_ a store because the store won't be created until `useStore()` is called inside of `setup()`:


posva:
Thanks! Ideally this should be organized differently:

- Add a cookbook entry about setup stores that describes everything about them from advantages, examples using composition api like watch, etc.
- Link to that page from https://pinia.vuejs.org/introduction.html#basic-example and the example you added

kawazoe:
> [...] from advantages, examples using composition api like watch, etc.

As a new user, here's what I would really like to see in the composition API docs:
- What is the proper way to declare the state? Should we use multiple refs, or a single reactive instance and toRefs?
- How should I access the state in actions? By using the ref directly or via this?
- Can I omit part of the state in the return object to make it private?
- Can part of the state be marked as readonly?
- Why is there no $patch/$reset methods on this in actions and is there a way to get a similar feature with this API?

Right now, some of that info is spread around issues and discussions. It would be nice for it to get centralized somewhere.

davidmeirlevy:
@kawazoe you're commenting on my actual PR and not on an issue.
Your comment is good and I'm afraid it will be lost after this PR will be closed.

kawazoe:
@davidmeirlevy I know, It was mostly as a continuation of @posva 's comment which suggested additions to the cookbook. I wasn't sure if the PR would get merged without those changes, so I decided to chip in anyway. :wink:

posva:
The PR won’t get merged without the changes indeed. If anybody wants to take it from here, feel free to do so!

packages/docs/core-concepts/index.md
```diff
@@ -14,6 +14,21 @@ export const useStore = defineStore('main', {
...
 
+Note that you can also define a store using a setup function instead, similar to Vue's Composition API:
+
+```js
+import { defineStore } from 'pinia'
+
+// useStore could be anything like useUser, useCart
+// the first argument is a unique id of the store across your application
+export const useStore = defineStore('main', () => {
+  return {
```
PatrikBird, line 25: I can't get it working with the return keyword in that spot... always used it like in https://github.com/antfu/vitesse/blob/main/src/stores/user.ts

davidmeirlevy: It works.
You probably didn't wrap your state with "toRefs" so you lost the reactivity.

PR #1025, opened 2022-02-03
The getters are actually receiving the 'store'

diegolepore:
My suggestion: I think the name of the parameter shouldn't be named 'state', since it is actually the entire 'store' object itself and not only the 'state'.


posva:
Thanks but this is intentional, most getters should only use the state yet having access to the store is more flexible but not typable. You should use `this.` when accessing the store.

PR #1032, opened 2022-02-04
docs: fix dark mode flashing bug

### LINKED ISSUE #686: docs page flickering
### Reproduction

Using latest chrome on a mac, open 2 tabs to https://pinia.esm.dev/  and the page starts flickering weirdly.

I can see it's caused by the dark mode flipping on and off rapidly.
I assume it has something to do with how dark mode selection is being saved in localstorage.

theoephraim:
Addresses #686

Currently multiple having multiple browser tabs open to the docs site creates some crazy flickering.

Also reloading the page creates inconsistent behaviour switching between dark/light mode.

This removes a bunch of extra code that was trying to mess with dark mode on page load that from what I can tell, seems unnecessary.

I believe the goal was to get the SSG of the site to include a "light" class on the body tag, but it appears from the current docs site page source that it wasn't actually working anyway...

Also
- now showing the dark/light mode toggle in the bottom left corner of the page
- removed what seemed to be unnecessary and broken copying of the pinia bundle into the docs site assets


packages/docs/package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "docs": "vitepress dev .",
     "docs:api": "node run-typedoc.js",
-    "docs:build": "vitepress build ."
+    "docs:build": "vitepress build .",
+    "docs:serve": "vitepress serve . --port 5001"
```
theoephraim: latest mac os uses port 5000 (the default) for airplay... 😠

packages/docs/vite.config.ts
```diff
@@ -28,23 +28,23 @@ export default defineConfig({
...
 
-function copyPiniaPlugin(): Plugin {
-  return {
-    name: 'copy-pinia',
-    async generateBundle() {
-      const filePath = path.resolve(
-        __dirname,
-        '../pinia/dist/pinia.esm-bundler.js'
-      )
+// function copyPiniaPlugin(): Plugin {
```
theoephraim: this copied file does not seem to be used anywhere...?

Regardless it was not working, and it appears the actual file is `../pinia/dist/pinia.esm-browser.js`

packages/docs/vite.config.ts
```diff
@@ -8,7 +8,7 @@ const fs = _fs.promises
 export default defineConfig({
   clearScreen: false,
   plugins: [
-    ...(process.env.NETLIFY ? [] : [copyPiniaPlugin()]),
+    // ...(process.env.NETLIFY ? [] : [copyPiniaPlugin()]),
```
theoephraim, line 11: it would seem the copy plugin shouldn't be triggered anyway locally due to the netlify check, but it was and it was failing.

theoephraim:
@plitter - fixed!

packages/docs/package.json
posva: Let’s not add a new script

posva (approved):
Thanks a lot! I added a few final touches to fix the initial theme

ghost:
I still see it

https://user-images.githubusercontent.com/11320080/154935113-f9039d41-3485-4853-932d-92ce4bf6ea26.mp4

theoephraim:
@web2033 - have you confirmed its not a cached version of the page?

ghost:
Yes.
For me, it flickers when in auto (light) mode.
In dark mode, it works without flickering.

PR #1040, opened 2022-02-09
docs: add hint on destructuring actions from store

taist24:
Added a warning section to hint the user on how to destructure `actions` from the `store`.

### DIFF
--- a/packages/docs/core-concepts/index.md
+++ b/packages/docs/core-concepts/index.md
@@ -84,8 +84,7 @@ export default defineComponent({
 })
 ```
 
-:::warning
-Extracting `actions` using `storeToRefs` will result in `actions` being `undefined`. To extract `actions` from the store, you should skip using `storeToRefs`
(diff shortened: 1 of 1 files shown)


posva (changes requested):
Let's add more code to the example above with an `increment` action and a comment above it

posva (changes requested):
I meant adding an example at https://github.com/vuejs/pinia/pull/1040/files#diff-8a9cb33bb1a330b9122d09d866b65789f9b8e62d1c02abaeb79ee9a1fb5a249cR77 instead of the warning you added, just add a new line showing how to extract the increment action and add a comment above it.

taist24:
Alright, and do you suggest keeping the "tip" section at the bottom? or removing it?

posva:
remove the whole warning

taist24:
Hey, were you able to review the last changes?  I've removed the warning layout and instead added it as an example.

posva:
Thanks, I did a few changes but couldn't push to your branch likely because you didn't check the "allow maintainers to edit"

PR #1044, opened 2022-02-10
docs: clarify import store requirement in Options API usage

bencodezen:
I was going through the docs and reviewing the Options API recommendations and realized there's a little bit of a disconnect between when stores are imported. So, rather than assume people can make the connection that the store is imported, I've added an explicit example to improve users' ability to figure out what comes from where.

### DIFF
--- a/packages/docs/core-concepts/getters.md
+++ b/packages/docs/core-concepts/getters.md
@@ -162,15 +162,29 @@ export default {
 You can use the same `mapState()` function used in the [previous section of state](./state.md#options-api) to map to getters:
 
 ```js
+// Example File Path:
+// src/stores/counterStore.js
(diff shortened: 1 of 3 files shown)


packages/docs/core-concepts/getters.md
```diff
@@ -107,9 +106,7 @@ export default {
 }
 </script>
 
-<template>
-User 2: {{ getUserById(2) }}
-</template>
+<template>User 2: {{ getUserById(2) }}</template>
```
bencodezen:
```suggestion
<template>
User 2: {{ getUserById(2) }}
</template>
```

posva (approved):
Thanks!

PR #1066, opened 2022-02-15
Add example for accessing router in a store

0xgdr:
This PR adds an example for how to access the vue router once you have added it as a plugin.


posva:
Thanks but how to use properties is already shown at the very beginning of the plugins section

PR #1074, opened 2022-02-17
chore(docs): fix message typo

zuramai:
Fixed typo message


bencodezen (approved):
Good catch! Thanks for the PR!

PR #1086, opened 2022-02-21
docs: add code exemple for Vue 2 unit tests

guillaumeduliscouet:
Can we show in the documentation how `@pinia/testing` can be used with Vue 2 and Vue Test Utils 1 ?


posva (changes requested):
Thanks! I think we should just add a short note instead linking to a new section at the very bottom to set this up with Vue 2

guillaumeduliscouet:
I tried an update @posva, tell me if it was what you were thinking about (and fill free to modify before merging)

posva (approved):

PR #1125, opened 2022-03-08
... or session storage

vanillajonathan:

### DIFF
--- a/packages/docs/core-concepts/plugins.md
+++ b/packages/docs/core-concepts/plugins.md
@@ -7,7 +7,7 @@ Pinia stores can be fully extended thanks to a low level API. Here is a list of 
 - Add new methods to stores
 - Wrap existing methods
 - Change or even cancel actions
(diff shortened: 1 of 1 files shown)


packages/docs/core-concepts/plugins.md
```diff
@@ -7,7 +7,7 @@ Pinia stores can be fully extended thanks to a low level API. Here is a list of
 - Add new methods to stores
 - Wrap existing methods
 - Change or even cancel actions
-- Implement side effects like local storage
+- Implement side effects like [local storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) or [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)
```
```suggestion
- Implement side effects like [Local Storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage)
```

posva (approved):
Thanks, I think one example is enough!

PR #1134, opened 2022-03-10
Update hot-module-replacement.md

hipertracker:
`import meta.hot` is raising a JS syntax error. The `module.hot` syntax is working.

### DIFF
--- a/packages/docs/cookbook/hot-module-replacement.md
+++ b/packages/docs/cookbook/hot-module-replacement.md
@@ -14,7 +14,7 @@ const useAuth = defineStore('auth', {
 })
 
 // make sure to pass the right store definition, `useAuth` in this case.
-if (import.meta.hot) {
-  import.meta.hot.accept(acceptHMRUpdate(useAuth, import.meta.hot))
+if (module.hot) {
+  module.hot.accept(acceptHMRUpdate(useAuth, module.hot))
 }
 ```


posva:
This syntax is for Vite. What are you using?

hipertracker:
I am using Webpack.

posva:
What version? Can you add this as a note below instead?

hipertracker:
> 
I am using Webpack 4.46.0.

hipertracker:
But in another project which uses Quasar2 + Vite also 'import.meta.hot' does not work:(

image

posva:
that seems more like like a tsconfig misconfiguration: missing the `vite/client` in `types` in tsconfig.json. But that's on quasar + vite

posva:
Given that there is already a note about wepback right above, I think it's better to keep it the way it is. Thanks.

hipertracker:
Adding this code to `quasar.d.ts` file has solved the problem:

```typescript
interface ImportMeta {
  readonly hot?: {
    readonly data: any

    accept(): void
    accept(cb: (mod: any) => void): void
    accept(dep: string, cb: (mod: any) => void): void
    accept(deps: string[], cb: (mods: any[]) => void): void

    prune(cb: () => void): void
    dispose(cb: (data: any) => void): void
    decline(): void
    invalidate(): void

    on(event: string, cb: (...args: any[]) => void): void
  }
}
```

posva:
I meant you need this: https://vitejs.dev/guide/features.html#client-types

PR #1138, opened 2022-03-11
docs: document createSpy with examples

mykwillis:
As described in this issue https://github.com/vuejs/pinia/discussions/1096.

I hit the same problem when setting up with vitest.

### DIFF
--- a/packages/docs/cookbook/testing.md
+++ b/packages/docs/cookbook/testing.md
@@ -102,6 +102,14 @@ expect(store.someAction).toHaveBeenCalledTimes(1)
 expect(store.someAction).toHaveBeenLastCalledWith()
 ```
 
+If you are not using Jest (for example, you are using vitest), you will need to provide a [createSpy](https://pinia.vuejs.org/api/interfaces/pinia_testing.testingoptions.html#createspy) to `createTestingPinia()`, like:
+
+```
+import { fn } from 'vitest'
+//...
+plugins: [createTestingPinia({ createSpy: fn })]
+```
+
 Please note that if you are using Vue 2, `@vue/test-utils` requires a [slightly different configuration](#unit-test-components-vue-2).
 
 You can find more examples in [the tests of the testing package](https://github.com/vuejs/pinia/blob/v2/packages/testing/src/testing.spec.ts).
--- a/packages/docs/cookbook/testing.md
+++ b/packages/docs/cookbook/testing.md
@@ -66,7 +66,7 @@ beforeEach(() => {
 
 ## Unit testing components
 
-This can be achieved with `createTestingPinia()`. I haven't been able to write proper documentation for this yet but its usage can be discovered through autocompletion and the documentation that appears in tooltips.
+This can be achieved with `createTestingPinia()`, which returns a pinia instance designed for unit tests.
 
 Start by installing `@pinia/testing`:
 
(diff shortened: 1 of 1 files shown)


packages/docs/cookbook/testing.md
```diff
@@ -102,6 +102,14 @@ expect(store.someAction).toHaveBeenCalledTimes(1)
 expect(store.someAction).toHaveBeenLastCalledWith()
 ```
 
+If you are not using Jest (for example, you are using vitest), you will need to provide a [createSpy](https://pinia.vuejs.org/api/interfaces/pinia_testing.testingoptions.html#createspy) to `createTestingPinia()`, like:
```
posva: It actually works if you have `globals: true` with vitest. Let's:

- Add a new subsection about actions being mocked
- Say they are automatically mocked in jest and in vite with globals: true
- Show how to add your own createSpy function

posva (changes requested):

packages/docs/cookbook/testing.md
mykwillis: Hmm, I have this is my vite.config.ts:

```
    test: {
      globals: true,
      environment: 'happy-dom',
    } as VitestUserConfig,
```

let me investigate.

mykwillis: @posva I think a new version of @pinai/testing needs to created and published. The behavior with `vi` was changed/fixed in [this commit](https://github.com/vuejs/pinia/commit/44f0762f93bf6ea389dd47574cbf4172989f616b) 11 days ago, but it's not published.

mykwillis: Well maybe not. I see that [vi is just a reference to vitest](https://github.com/vitest-dev/vitest/blob/c1fa19dfc2ccf6c27b2289dc829ec29c53ec66b6/packages/vitest/src/integrations/utils.ts#L86), so either of those symbols should work. I'll have to find another reason why this isn't working on my machine.

(Incidentally, I can see the `vi` global in my test file, but still get the error when calling `createTestingPinia`.

mykwillis: ok @posva I'm sorry for all the noise here, but yes, the reason that you still need to supply `createSpy` with the released version of @pinia/testing (0.0.9) is that support for detecting `vi`/`vitest` was [only added 11 days ago,](e4638ef34c16487b2f35b0de09800f6dcfd5a980) and there has not been a release of @pinia/testing since then.

I'm unfamiliar with your release schedule so perhaps that is intentional. 

I'll prepare this PR to update the documentation in anticipation of a new release being available (it looks like there are comments in the code along the lines of what you'd like the docs to say so I'll use those as a guide).

posva: Sure! I probably haven’t released because there are a few things to change before

packages/docs/cookbook/testing.md
```diff
@@ -66,7 +66,7 @@ beforeEach(() => {
 
 ## Unit testing components
 
-This can be achieved with `createTestingPinia()`. I haven't been able to write proper documentation for this yet but its usage can be discovered through autocompletion and the documentation that appears in tooltips.
+This can be achieved with `createTestingPinia()`, which returns a pinia instance designed for unit tests.
```
posva:
```suggestion
This can be achieved with `createTestingPinia()`, which returns a pinia instance designed to help unit tests components.
```

packages/docs/cookbook/testing.md
```diff
@@ -94,23 +94,54 @@ store.name = 'my new name'
 store.$patch({ name: 'new name' })
 expect(store.name).toBe('new name')
 
-// actions are stubbed by default but can be configured by
-// passing an option to `createTestingPinia()`
+// actions are stubbed by default, so this call won't actually
+// execute the implementation defined by the store. See below 
+// to customize this behavior.
```
posva:
```suggestion
// actions are stubbed by default, meaning they don't execute their code by default.
// See below to customize this behavior.
```

packages/docs/cookbook/testing.md
```diff
@@ -94,23 +94,54 @@ store.name = 'my new name'
...
 expect(store.someAction).toHaveBeenLastCalledWith()
 ```
 
-If you are not using Jest (for example, you are using vitest), you will need to provide a [createSpy](https://pinia.vuejs.org/api/interfaces/pinia_testing.testingoptions.html#createspy) to `createTestingPinia()`, like:
+Please note that if you are using Vue 2, `@vue/test-utils` requires a [slightly different configuration](#unit-test-components-vue-2).
+
+
+### Customizing behavior of actions
+
+By default, `createTestingPinia` stubs out all store actions, such that they are not executed when called during unit tests. This allows you to test your components and stores separately.
```
posva:
```suggestion
`createTestingPinia` stubs out all store actions unless told otherwise. This allows you to test your components and stores separately.
```

packages/docs/cookbook/testing.md
```diff
@@ -94,23 +94,54 @@ store.name = 'my new name'
...
+
+If you want your actions to execute during unit tests, specify `stubActions: true` as an option when calling `createTestingPinia`:
```
posva:
```suggestion
If you want to revert this behavior and normally execute your actions during tests, specify `stubActions: false` when calling `createTestingPinia`:
```

packages/docs/cookbook/testing.md
```diff
@@ -94,23 +94,54 @@ store.name = 'my new name'
...
 
 ```
```
posva:
```suggestion
```js
```

packages/docs/cookbook/testing.md
```diff
@@ -94,23 +94,54 @@ store.name = 'my new name'
...
+### Specifying the createSpy function
+
+When using Jest, or vitest with `globals: true`, `createTestingPinia` will create action stubs automatically using the spy function defined by the test framework (`jest.fn` or `vitest.fn`). If you are using a different framework, you'll need to provide a [createSpy](https://pinia.vuejs.org/api/interfaces/pinia_testing.testingoptions.html#createspy) and supply it as a parameter.
+
+```
+import sinon from 'sinon'
+createTestingPinia({ 
+  createSpy: sinon.spy // use sinon's spy to wrap actions
+  })
+```
```
posva:
```suggestion
```js
import sinon from 'sinon'

createTestingPinia({ 
  createSpy: sinon.spy // use sinon's spy to wrap actions
})
```
```

File: packages/docs/cookbook/testing.md [hunk 8]
```diff
@@ -94,23 +94,54 @@ store.name = 'my new name'
...
+
+// ...but it's still wrapped with a spy, so you can inspect calls 
+expect(store.someAction).toHaveBeenCalledTimes(1)
+
 ```
 
-Please note that if you are using Vue 2, `@vue/test-utils` requires a [slightly different configuration](#unit-test-components-vue-2).
+### Specifying the createSpy function
+
+When using Jest, or vitest with `globals: true`, `createTestingPinia` will create action stubs automatically using the spy function defined by the test framework (`jest.fn` or `vitest.fn`). If you are using a different framework, you'll need to provide a [createSpy](https://pinia.vuejs.org/api/interfaces/pinia_testing.testingoptions.html#createspy) and supply it as a parameter.
```
posva:
```suggestion
When using Jest, or vitest with `globals: true`, `createTestingPinia` automatically stubs actions using the spy function based on the existing test framework (`jest.fn` or `vitest.fn`). If you are using a different framework, you'll need to provide a [createSpy](/api/interfaces/pinia_testing.testingoptions.html#createspy) option:
```

packages/docs/cookbook/testing.md
```diff
@@ -94,23 +94,54 @@ store.name = 'my new name'
...
+
+```
+import sinon from 'sinon'
+createTestingPinia({ 
+  createSpy: sinon.spy // use sinon's spy to wrap actions
+  })
+```
+
+(Note that the provided `createSpy` will also be used to wrap the `$patch` method of the store).
+
```
posva:
```suggestion
```

posva (approved):
Thanks for the help with this!

PR #1150, opened 2022-03-15
docs: update nuxt.js examples in plugins.md

theolavaux:
The `myPlugin` function was missing an arrow.
I also updated the way the `creationTime` property was added to the global state.

### DIFF
--- a/packages/docs/core-concepts/plugins.md
+++ b/packages/docs/core-concepts/plugins.md
@@ -371,11 +371,8 @@ function MyPiniaPlugin({ store }: PiniaPluginContext) {
     console.log(`[🍍 ${mutation.storeId}]: ${mutation.type}.`)
   })
 
-  const creationTime = new Date();
-   
-  // add new state
-  store.$state.creationTime = creationTime;
-  store.secret.creationTime = creationTime;
+  // Note this has to be typed if you are using TS
+  return { creationTime: new Date() }
 }
 
 const myPlugin: Plugin = ({ $pinia }) => {
--- a/packages/docs/core-concepts/plugins.md
+++ b/packages/docs/core-concepts/plugins.md
@@ -371,12 +371,17 @@ function MyPiniaPlugin({ store }: PiniaPluginContext) {
     console.log(`[🍍 ${mutation.storeId}]: ${mutation.type}.`)
   })
 
-  return { creationTime: new Date() }
+  const creationTime = new Date();
+   
+  // add new state
+  store.$state.creationTime = creationTime;
+  store.secret.creationTime = creationTime;
 }
 
-const myPlugin: Plugin = ({ pinia }) {
-  pinia.use(MyPiniaPlugin);
+const myPlugin: Plugin = ({ $pinia }) => {
+  $pinia.use(MyPiniaPlugin);
 }
+
 export default myPlugin
 ```
 
--- a/packages/docs/core-concepts/plugins.md
+++ b/packages/docs/core-concepts/plugins.md
@@ -376,7 +376,7 @@ function MyPiniaPlugin({ store }: PiniaPluginContext) {
 }
 
 const myPlugin: Plugin = ({ $pinia }) => {
-  $pinia.use(MyPiniaPlugin);
+  $pinia.use(MyPiniaPlugin)
 }
 
 export default myPlugin


packages/docs/core-concepts/plugins.md
```diff
@@ -365,18 +365,23 @@ When [using pinia alongside Nuxt](../ssr/nuxt.md), you will have to create a [Nu
 import { PiniaPluginContext } from 'pinia'
 import { Plugin } from '@nuxt/types'
 
-function MyPiniaPlugin({ store }: PiniaPluginContext) {
+const MyPiniaPlugin = ({ store }: PiniaPluginContext) => {
```
posva: Let's keep the previous syntax

packages/docs/core-concepts/plugins.md
```diff
@@ -365,18 +365,23 @@ When [using pinia alongside Nuxt](../ssr/nuxt.md), you will have to create a [Nu
...
     // react to store changes
     console.log(`[🍍 ${mutation.storeId}]: ${mutation.type}.`)
   })
 
-  return { creationTime: new Date() }
+  const creationTime = new Date();
+   
+  // add new state
+  store.$state.creationTime = creationTime;
+  store.secret.creationTime = creationTime;
```
posva: Why are you adding this to state and secret? Returning it as properties was good enough for the purpose of the example. I think we should keep the previous version.

posva (changes requested):
Thanks!

packages/docs/core-concepts/plugins.md
posva: can you revert this?

theolavaux: My bad, I thought v2 was the live branch for the docs, there is a syntax error in the live version (missing arrow in arrow function https://pinia.vuejs.org/core-concepts/plugins.html#nuxt-js)

Screenshot 2022-03-15 at 20 11 30

packages/docs/core-concepts/plugins.md
theolavaux: I should have included more context you're right. By just copy/pasting the example, I get a Typescript error on `$pinia.use(MyPiniaPlugin);`

```
Argument of type '({ store }: PiniaPluginContext<string, StateTree, _GettersTree<StateTree>, _ActionsTree>) => { creationTime: Date; }' is not assignable to parameter of type 'PiniaPlugin'.
  Type '{ creationTime: Date; }' is not assignable to type 'void | Partial<PiniaCustomProperties<string, StateTree, _GettersTree<StateTree>, _ActionsTree> & PiniaCustomStateProperties<...>>'.ts(2345)
```

When I add the property to the state/secret instead of returning it, the error goes away. Is there a difference between adding to those two and the return ?

packages/docs/core-concepts/plugins.md
theolavaux: Is there a technical reason for it ?

packages/docs/core-concepts/plugins.md
posva: In order to make the typing work, you need to add the TS definitions as shown in the Plugin section

packages/docs/core-concepts/plugins.md
```diff
@@ -371,12 +371,17 @@ function MyPiniaPlugin({ store }: PiniaPluginContext) {
...
```
```suggestion
  // Note this has to be typed if you are using TS
  return { creationTime: new Date() }
```

posva (approved):
Thanks!

packages/docs/core-concepts/plugins.md
```diff
@@ -371,12 +371,14 @@ function MyPiniaPlugin({ store }: PiniaPluginContext) {
...
   })
 
+  // Note this has to be typed if you are using TS
   return { creationTime: new Date() }
 }
 
-const myPlugin: Plugin = ({ pinia }) {
-  pinia.use(MyPiniaPlugin);
+const myPlugin: Plugin = ({ $pinia }) => {
+  $pinia.use(MyPiniaPlugin);
```
```suggestion
  $pinia.use(MyPiniaPlugin)
```

packages/docs/core-concepts/plugins.md
theolavaux: I got confused because at the top of the file it says `// plugins/myPiniaPlugin.js` but the content is Typescript and it's also stated below `Note the above example is using TypeScript...`

PR #1165, opened 2022-03-24
fix(testing): Vue 2 initial state reactive

blake-newman:
Vue 2 require observable state in order to detect changes; the initial state needs to explicitly needs to be made reactive so changes will be observed in components.

### DIFF
--- a/packages/testing/src/testing.ts
+++ b/packages/testing/src/testing.ts
@@ -1,10 +1,11 @@
-import { App, createApp } from 'vue-demi'
+import { App, createApp, isReactive, isRef, isVue2, set } from 'vue-demi'
 import {
   Pinia,
   PiniaPlugin,
   setActivePinia,
   createPinia,
   StateTree,
+  _DeepPartial,
 } from 'pinia'
 
 export interface TestingOptions {
@@ -89,7 +90,7 @@ export function createTestingPinia({
 
   pinia.use(({ store }) => {
     if (initialState[store.$id]) {
-      store.$patch(initialState[store.$id])
+      mergeReactiveObjects(store.$state, initialState[store.$id])
     }
   })
 
@@ -130,4 +131,45 @@ export function createTestingPinia({
   })
 
   return pinia as TestingPinia
+}
+
+function mergeReactiveObjects<T extends StateTree>(
+  target: T,
+  patchToApply: _DeepPartial<T>
+): T {
+  // no need to go through symbols because they cannot be serialized anyway
+  for (const key in patchToApply) {
+    const subPatch = patchToApply[key]
+    const targetValue = target[key]
+    if (
+      isPlainObject(targetValue) &&
+      isPlainObject(subPatch) &&
+      !isRef(subPatch) &&
+      !isReactive(subPatch)
+    ) {
+      target[key] = mergeReactiveObjects(targetValue, subPatch)
+    } else {
+      if (isVue2) {
+        set(target, key, subPatch)
+      } else {
+        // @ts-expect-error: subPatch is a valid value
+        target[key] = subPatch
+      }
+    }
+  }
+
+  return target
+}
+
+function isPlainObject<S extends StateTree>(value: S | unknown): value is S
+function isPlainObject(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  o: any
+): o is StateTree {
+  return (
+    o &&
+    typeof o === 'object' &&
+    Object.prototype.toString.call(o) === '[object Object]' &&
+    typeof o.toJSON !== 'function'
+  )
 }


posva:
Since we are using `patch` this shouldn't be necessary 🤔 Do you have a repro or a failing test case (inside pinia repro) that I can take a look at?

blake-newman:
@posva after digging deeper I realised that is affects when there is partial state. This state structure occurs when data may not exist until it's fetched from an API.

In tests `initialState` would be used to initiate the data without the need for fetching or interacting with the internals for the store (calling actions etc). 

I couldn't get this working with stackblitz; as there is vue-demi issue occurring. However here is a zip of a reproduction.

[pinia-vue2-example.zip](https://github.com/vuejs/pinia/files/8343592/pinia-vue2-example.zip)

packages/testing/src/testing.ts
```diff
@@ -86,10 +86,11 @@ export function createTestingPinia({
   createSpy: _createSpy,
 }: TestingOptions = {}): TestingPinia {
   const pinia = createPinia()
+  const state = isVue2 ? Vue2.observable(initialState) : initialState
 
   pinia.use(({ store }) => {
-    if (initialState[store.$id]) {
-      store.$patch(initialState[store.$id])
+    if (state[store.$id]) {
+      store.$patch(state[store.$id])
```
posva: I think we should use a custom patch function that calls `$set()` in Vue 2. The patch is originally using this function in `store.ts`

```ts
function mergeReactiveObjects<T extends StateTree>(
  target: T,
  patchToApply: _DeepPartial<T>
): T {
  // no need to go through symbols because they cannot be serialized anyway
  for (const key in patchToApply) {
    const subPatch = patchToApply[key]
    const targetValue = target[key]
    if (
      isPlainObject(targetValue) &&
      isPlainObject(subPatch) &&
      !isRef(subPatch) &&
      !isReactive(subPatch)
    ) {
      target[key] = mergeReactiveObjects(targetValue, subPatch)
    } else {
      // @ts-expect-error: subPatch is a valid value
      target[key] = subPatch
    }
  }

  return target
}
```

I think that for testing purposes, we could call `$set()` only in Vue 2 instead of doing `target[key] = ...`. Can you copy that function and add it to the testing package?

posva (changes requested):

packages/testing/src/testing.ts
blake-newman: @posva 

I noticed the `hmr` `patchObject` method handles reactivity for Vue2. 

```
export function patchObject(
  newState: Record<string, any>,
  oldState: Record<string, any>
): Record<string, any> {
  // no need to go through symbols because they cannot be serialized anyway
  for (const key in oldState) {
    const subPatch = oldState[key]

    // skip the whole sub tree
    if (!(key in newState)) {
      continue
    }

    const targetValue = newState[key]
    if (
      isPlainObject(targetValue) &&
      isPlainObject(subPatch) &&
      !isRef(subPatch) &&
      !isReactive(subPatch)
    ) {
      newState[key] = patchObject(targetValue, subPatch)
    } else {
      // objects are either a bit more complex (e.g. refs) or primitives, so we
      // just set the whole thing
      if (isVue2) {
        set(newState, key, subPatch)
      } else {
        newState[key] = subPatch
      }
    }
  }

  return newState
}
```

Perhaps we should do the same thing for `mergeReactiveObjects` not just in test code. Otherwise it may cause discrepancies in behaviour between testing, HMR and application code. It a nice DX for the `$patch` to consider reactivity principles of Vue 2 and avoid the need for users to consider the reactivity constraints (supporting migration from Vue 2 & Vue 3). Thoughts?

posva: The three cases are really different: HMR (adding, removing new state) and initialState (edge case you highlighted) might need set but usually patch in application code doesn’t (same caveats as Vue, so consistent).

blake-newman: @posva fair enough. I pushed changes. `isPlainObject` was not exposed i assume by design so duplicated that method in the testing package also.

packages/testing/src/testing.ts
```diff
@@ -89,7 +90,7 @@ export function createTestingPinia({
 
   pinia.use(({ store }) => {
     if (initialState[store.$id]) {
-      store.$patch(initialState[store.$id])
+      mergeReactiveObjects(store.$state[store.$id], initialState[store.$id])
```
posva: didn't you mean

```suggestion
      mergeReactiveObjects(store.$state, initialState[store.$id])
```

posva (changes requested):

packages/testing/src/testing.ts
blake-newman: I don't think so - wouldn't this be

```
const $state = {
   module: {
     isBoolean: true
   }
}

const initialState = {
  module: {
     isBoolean: false
   }
}

mergeReactiveObjects($state['module'], initialState['module'])

/* $state = {
  module: {
     isBoolean: false
   }
} */
```

Where as the suggestion looks like it would do:

```ts
mergeReactiveObjects($state, initialState['module'])

/* $state = {
  module: {
     isBoolean: true
   },
   isBoolean: false
} */
```

blake-newman: Ah wait so yes - forgot each store is not nested like Vuex :D

posva: initialState is meant to be the rootState while `store.$state` is a store's local state

posva (approved):
Thanks!

PR #1166, opened 2022-03-24
docs: showing how to use initialState in testing

Flavsditz:
When looking at how to properly make use of the initialState I found the documentation lacking.
With the help of the community, I found the response and I wanted to give back by enriching the docs for the next person.


posva (approved):
Thanks, I simplified the example a bit and ended up rewriting the whole thing 😅

Flavsditz:
Haha no problem @posva, as long as the message gets across I am good ;)

PR #1169, opened 2022-03-25
docs: typo

dwarner5:
Fixed typos: 
-cleanup up
+cleaned up

Doc update only.


posva (changes requested):
This is actually generated, can you modify it in the source code instead?

dwarner5:
I've updated the doc source comment in the types.ts file, but I'm unable to successfully build the docs. I've tried 'npm run docs:api' from the packages/docs folder, but I'm not familiar with how that is supposed to work. Apologies if not following the proper procedure. Any guidance would be well received.

posva (approved):
Thanks! No need to generate them

PR #1174, opened 2022-03-27
Docs: Add section to show using a store outside Vue.

cwilvx:
This pull request adds a section to show how to use a Pinia store outside a Vue file to the Pinia docs.

### DIFF
--- a/packages/docs/core-concepts/outside-component-usage.md
+++ b/packages/docs/core-concepts/outside-component-usage.md
@@ -52,6 +52,33 @@ router.beforeEach((to) => {
 })
 ```
 
+## Using pinia outside Vue
+To use a store outside a Vue instance, you will have to define your Pinia store inside your Vue file and pass it outside Vue as an argument. For example, inside your Vue file you can do something like this:
+
+```js
+import { createRouter } from 'vue-router'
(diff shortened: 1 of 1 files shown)


posva:
What is this adding? I don't understand what new information it brings

cwilvx:
@posva  

Under [Using a store outside of a component](https://pinia.vuejs.org/core-concepts/outside-component-usage.html), the Pinia docs just mention that ` ... if the pinia instance cannot be automatically injected, you have to manually provide it to the useStore() function` but doesn't show how to actually do so. 

This pull request adds an example to demonstrate how to inject a Pinia store outside Vue.

posva:
Yeah, that's explained in the SSR docs

cwilvx:
@posva 

What about client side rendered apps? Some people don't use SSR, they expect to find the info on the `Using a store outside of a component` page.

posva:
Your example is just showing how to pass a store as an argument of a function. Are you trying to show anything more?

cwilvx:
It's just showing passing a store as an argument as a way of injecting a store outside a vue instance. Nothing more.

posva:
I see. Thank you but I don't think this should be in the docs.

PR #1186, opened 2022-04-01
Explaint how to data-bind pinia to form inputs

DespertaWeb:
This topic is super important and it was treated not specifically. Now we've got and example


posva:
Thanks but I don’t think showing this makes things clearer. After all using v mode with pink just works

PR #1201, opened 2022-04-04
fix(testing): allow writable pinia getters in Vue 2

### LINKED ISSUE #1200: WritableComputed doesn't find getters for Vue2 testing
### Reproduction

https://github.com/heykc/pinia-test

1. pull down the repo above
2. run the test
3. notice test fails with a warning of `[Vue warn]: Write operation failed: computed value is readonly.`

### Expected behavior

The test should successfully reassign the `double` getter to a new value, and the expectation following should pass.

This test is nearly identical to the test inside of @pinia/testing called `allows overriding computed properties` but runs with a Vue 2 instance instead.

### Actual behavior

The test fails with a warning of `[Vue warn]: Write operation failed: computed value is readonly.`

### Additional information

I believe this is due to the fact that the `WritableComputed` function in `@pinia/testing/testing.ts` only looks for refs when checking for computed properties. In Vue 2, there are no refs and therefore none of the getters within a given store are spotted within this function. 

I have a proposed solution which is to check for Vue2 specifically and confirm there are getters within the options of the store. The code below is an example of my solution: 

```js
function WritableComputed({ store, options }: PiniaPluginContext) {
  const rawStore = toRaw(store)
  for (const key in rawStore) {
    const value = rawStore[key]
    // Vue 2 pinia getters are not refs and have to be found a different way than isComputed()
    const isVue2Getter = isVue2 && options.getters?.[key]
    if (isComputed(value) || isVue2Getter) {
      rawStore[key] = customRef((track, trigger) => {
        let internalValue: any
        return {
          get: () => {
            track()
            // since Vue2 getters are just functions and not refs, they don't have a value property
            return internalValue !== undefined ? internalValue : (isVue2Getter ? value : value.value)
          },
          set: (newValue) => {
            internalValue = newValue
            trigger()
          },
        }
      })
    }
  }
}
```
I'd be happy to put in a pull request but I couldn't figure out a way to reproduce this test directly in the pinia/testing repo. It would have required a way to switch `vue-demi` to vue2 which caused a lot of peer dependency conflicts. I'd love to find out the best way to write a test for this particular instance to go along with this code fix.

I'm also not married to the isVue2Getter variable name or the nested ternary. Open to code suggestions on this.

heykc:
## Description:
This allows Pinia store getters in Vue 2 apps to be reassigned as shown in the [test here](https://github.com/vuejs/pinia/blob/6ce186f5acf50c0c4ece1aec6468534e0916e688/packages/testing/src/testing.spec.ts#L203). 

## Related Issues:
Close #1200 

## Testing Notes:
I wasn't able to write a test for this directly into the pinia/testing repo due to the fact that I would need to use the `vue-demi switch 2 vue2` npm script, which requires me to install vue/composition api, which then causes a number of peer dependency issues. I talk about this further in the Issue attached.

## Code Notes:
I'm not married to the nested ternary, and I don't know if there's a precedence for optional chaining, so I'm happy to adjust any of these.

### DIFF
--- a/packages/testing/src/testing.ts
+++ b/packages/testing/src/testing.ts
@@ -197,17 +197,19 @@ function isComputed<T>(
   return !!v && isRef(v) && 'effect' in v
 }
 
-function WritableComputed({ store }: PiniaPluginContext) {
+function WritableComputed({ store, options }: PiniaPluginContext) {
   const rawStore = toRaw(store)
   for (const key in rawStore) {
     const value = rawStore[key]
-    if (isComputed(value)) {
+    // Vue2 store getters are not refs and can't be found with isComputed()
+    const isVue2Getter = isVue2 && options.getters?.[key]
+    if (isComputed(value) || isVue2Getter) {
       rawStore[key] = customRef((track, trigger) => {
         let internalValue: any
         return {
           get: () => {
             track()
-            return internalValue !== undefined ? internalValue : value.value
+            return internalValue !== undefined ? internalValue : (isVue2Getter ? value : value.value)
           },
           set: (newValue) => {
             internalValue = newValue


packages/testing/src/testing.ts
```diff
@@ -197,17 +197,19 @@ function isComputed<T>(
...
+    // Vue2 store getters are not refs and can't be found with isComputed()
+    const isVue2Getter = isVue2 && options.getters?.[key]
+    if (isComputed(value) || isVue2Getter) {
       rawStore[key] = customRef((track, trigger) => {
         let internalValue: any
         return {
           get: () => {
             track()
-            return internalValue !== undefined ? internalValue : value.value
+            return internalValue !== undefined ? internalValue : (isVue2Getter ? value : value.value)
```
posva, line 212: shouldn't this be
```suggestion
            return internalValue !== undefined ? internalValue : (isVue2Getter ? store[key] : value.value)
```

I haven't tested it but I don't think reading `value` will be reactive

heykc: @posva ah, you're right, the reactivity isn't retained with just value. though using `store[key]` causes an infinite get loop 😅 I think the correct move here would be to call the function from `options.getters`, no?

```suggestion
            return internalValue !== undefined ? internalValue : (isVue2Getter ? options.getters![key](store): value.value)
```

posva: Ah true about the infinite loop! The problem with options is that it’s not always there eg setup stores. I think there is an open issue in composition api regarding the detection of computed properties. We will likely need to push that forward instead

heykc: Hmmm I think I see what you're saying. Are you referring to someone using setup stores in vue 2? I wasn't aware that was possible. If that's the case then yeah, you're right. this wouldn't work for that particular scenario.

posva: Yes

heykc:
Hey @posva sorry about the radio silence, got busy these last couple weeks. Based on your feedback it sounds like we should close this PR but maybe leave the issue? Would you agree?

posva:
It’s up to you: it’s okay if you want to keep it open for later

packages/testing/src/testing.ts
heykc: @posva I was looking into the vue/composition-api plugin docs a little more and it looks like you merged a PR that does differentiate computed properties in composition. Since you're looking for `effect` in your `isComputed` function _and_ I'm looking for `options.getters?.[key]` in `isVue2Getter`, that tells me that all bases are covered here. `isVue2Getters` couldn't be true if there were no options.getters object in the store.

I'm going to add another test to the mock project I'm using to test vue2 with the vue/composition-api plugin to confirm, but does this sound right to you?

posva: The problem is `toRaw()` doesn't work in composition-api

MisterIsaak:
So is there a known workaround to mock getters in Vue2? So far I've been resorting to supplying the required state data to have the getters return as expected. But it's fairly verbose at times for complex getters.

rhysv96:
also interested in mocking getters in vue 2. if i mock the state then that means im also testing store code, which isnt ideal for component unit tests.

heykc:
This is how I've been getting around it IF you're writing getters in pinia with Options API. I mocked the utilization to be fairly similar to how initialState is structured (storeId key, an object of getters as value). Example below it. 

```js
/**
 * A Pinia plugin for mocking getters in component tests.
 *
 * @param {object} gettersMap
 * @returns {Function}
 */
export const initialGetters = (gettersMap) => ({ store, options }) => {
  if (gettersMap[store.$id]) {
    Object.keys(options.getters).forEach((getter) => {
      Object.defineProperty(store, getter, {
        get: () => gettersMap[store.$id][getter] || options.getters[getter],
      });
    });
  }
};
```

```js
// Example.spec.js

mount(Example, {
  pinia: createTestingPinia({
    plugins: [
      initialGetters({
        myStore: {
          someGetter: jest.fn(),
          anotherGetter: jest.fn(() => 12),
        },
      }),
    ],
  }),
});
```

Graphmaxer:
Any news ?

djodjonx:
here a workaround: https://github.com/vuejs/pinia/issues/1200#issuecomment-2373791316

Issue #981, opened 2022-01-23
Autocomplete/intellisense/types for "actions" in defineStore() typed as "any" (missing)


nurdism:
You can actually see this in the demo projects on [stackblitz](https://stackblitz.com/github/piniajs/example-nuxt-3?file=stores%2Fcounter.ts), when writing a func in actions, `this` is typed as `any` and doesn't provide any autocomplete/intellisense/types

![](https://i.imgur.com/m4AGnZu.png)

```ts
export const useCounter = defineStore({
  id: 'counter',
  state: () => ({
    n: 2,
  }),
  getters: {
    double: (state) => state.n * 2,
  },
  actions: {
    increment(amount = 1) {
      this.n += amount // this is "any", not typesafe
    },
  },
})
```

posva:
It was missing the `strict: true` in the TS config

Asarew:
even with `strict: true` the error is still there.

Bornegrim:
Try setting `noImplicitThis` to `true` in your TS config.

realityfilter:
I was hitting this problem after refactoring a mono repo. The original package had the `strict: true` setting while the new one had not. Why is this necessary? I think this should be in the docs. It is hard to find the origin of this configuration error.

dulimarta:
I had a similar issue in my TypeScript code and was able to get around it by typecasting the actions to `ThisType<MyAppState>` something like:

```ts
type MyAppState = {
  counter: number;
}


export const useCounter = defineStore("xyz", {
  state: (): MyAppState => ({
    counter: 0
  }),
  action: {
    increment() {
    }
  } as ThisType<MyAppState>
});
```

PR #1226, opened 2022-04-13
docs: minimum tsconfig.json props for proper TS support

ghost:
https://github.com/vuejs/pinia/issues/981

Perhaps putting it in the Nuxt section when it's related to TypeScript in general is not the best, but this is the only TS-related section in the docs afaik

### DIFF
--- a/packages/docs/ssr/nuxt.md
+++ b/packages/docs/ssr/nuxt.md
@@ -90,6 +90,6 @@ If you are using TypeScript or have a `jsconfig.json`, you should also add the t
 }
 ```
 
-Make sure to set `strict: true` or at least `noImplicitAny: true` to get proper typing and completions when using `this.` in store definitions
-
 This will also ensure you have autocompletion 😉 .
+
+Make sure to set `strict: true` or at least `noImplicitAny: true` to avoid typing of `this.` as `any` in store definitions.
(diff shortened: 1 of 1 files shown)


posva:
doesn't the default nuxt config has a strict: true ? I don't think this level of information should be included in docs because by default, any project should have strict: true in their tsconfig.

ghost:
@posva you're right that it's default. I could change "make sure to set" to "make sure you have".

Pinia can be installed in:
1. a fresh project with strict: true already there
2. in an already existing one (the existence of disableVuex proves this) with some nightmarish mangled tsconfig.json

There are *probably* a few already existing projects with strict: false which would like to use Pinia in the future. You can count the 6 people in the linked issue as some of them

There isn't really much of a hint on how to fix that issue from TypeScript itself - `this` is just any. I resorted to comparing the tsconfig.json where `this.` was working (create-vue) to my non-working tsconfig.json and modifying property-by-propety and reloading tsserver. Only when I searched "strict: true" I could find the linked issue. Hence my PR. Feel free to close if you don't agree, this is just one line of text after all :)

posva:
Yeah, I don' think this should be in the docs, at least for the moment. It should also not be in the nuxt section. Thanks though

PR #1232, opened 2022-04-17
Fix typo in migration-vuex.md

rhorber:
Add missing `d`.


posva:
Thanks but this is actually correct.

rhorber:
After reading it again: You're absolutely right. I'm sorry, I read it from a wrong perspective (and only from that).

PR #1235, opened 2022-04-19
refactor: more concise code， isListening == false to !isListening

vaebe:
more concise code， isListening == false to !isListening。


posva:
Thanks but this is intentional, the value starts as undefined

PR #1244, opened 2022-04-21
docs: fix typo in `PiniaPluginContext`

lazzzis:
`options` is an object defining the store passed to `defineStore()`, per https://pinia.vuejs.org/core-concepts/plugins.html#introduction

### DIFF
--- a/packages/pinia/src/rootStore.ts
+++ b/packages/pinia/src/rootStore.ts
@@ -122,7 +122,7 @@ export interface PiniaPluginContext<
   store: Store<Id, S, G, A>
 
   /**
-   * Current store being extended.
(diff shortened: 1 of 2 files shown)


packages/pinia/src/rootStore.ts
```diff
@@ -122,7 +122,7 @@ export interface PiniaPluginContext<
   store: Store<Id, S, G, A>
 
   /**
-   * Current store being extended.
+   * Options object defining the store passed to `defineStore()`.
```
posva:
```suggestion
   * Initial options defining the store when calling `defineStore()`.
```

posva (approved):
Thanks!

PR #1280, opened 2022-05-08
test: polish 

lazzzis:
1. remove a test case "throws errors" which seems duplicated
2. fix some typos and remove some comments which seem out-of-date
3. test the getter `sum` as well in the test case "keeps getters reactive"


packages/pinia/__tests__/getters.spec.ts
```diff
@@ -151,16 +151,21 @@ describe('Getters', () => {
       expectType<() => number>(b.incrementA)
 
       expect(a.double).toBe(0)
+      expect(a.sum).toBe(0)
```
posva: can you remove these checks of a.sum? They shouldn't be in this test

posva (changes requested):
Thanks!

posva (approved):

PR #1294, opened 2022-05-14
feat: allow $onAction callbacks to change behavior

shotwn:
## Description
Currently action.onError() is not very usable. If user wants to handle the error globally using onError(), error has to propagate out of action. This will trigger onError() but there is no way to prevent error to propagate further on and user has a unhandled error warning. If error is to be handled within action, onError() doesn't get triggered.

Researching a solution I noticed a 2 lines in API docs.
> after(callback): void 
> ...Can return a value (other than undefined) to override the returned value.

> onError(callback): void 
> ...Return false to catch the error and stop it from propagating.

## Implementation
- action.onError(): Can stop error propagation if callback returns false.
- action.after() : Can override action's return value.
- - Only one of the callbacks is allowed to do override. More than one will cause a console warning.
- - Last result in array of results will be used for override.

### DIFF
--- a/packages/docs/core-concepts/actions.md
+++ b/packages/docs/core-concepts/actions.md
@@ -202,6 +202,10 @@ const unsubscribe = someStore.$onAction(
           Date.now() - startTime
         }ms.\nResult: ${result}.`
       )
+
+      // returning a value will override action's return value.
+      // this should be done only once.
+      return result * 2
     })
 
     // this will trigger if the action throws or returns a promise that rejects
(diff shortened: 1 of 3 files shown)


posva:
Thanks but this was intentionally removed at 4dc1f1bec8f3db508c02fa6bdcc8203f280cbb3e. I updated the docs instead.

I think to handle errors globally, there should be an RFC that takes into account the comments in the commit above. This could maybe be a new plugin too that enables new function on `pinia` like `pinia.onError()`.

Overriding action returns has proven to be problematic so far

PR #1309, opened 2022-05-19
docs(en): use case for option function in defineStore

trchopan:


posva:
This example could be useful for a cookbook entry

trchopan:
Thank you for your response. I've changed it to be a cookbook entry.

packages/docs/cookbook/create-store-dynamically.md
```diff
@@ -0,0 +1,51 @@
...
+
+Store can be created dynamically using factory pattern. Which will greatly help reduce boilerplate and structure your application.
+
+## Example
+
+An usecase of multiple tables that require pagination and filter. And all need to have separate stores to keep track of the result and pagings.
+
+We can have a creator function like this
+
+```js
```
Ilanaya, line 11: In this example, there should be `ts` if you want to use type annotations. Nothing critical, but the formatting would be a little bit better

Ilanaya (commented):
By the way, this pr looks misleading to me - I was wandering around to find some solution for dynamic store registration (implying passing different store `ids` to the factory function) and got here.

This pr, as I can see, better explains how stores handle states rather than how to create them dynamically. I think the contents should be refactored to better match the pr point.

posva:
Thanks for the PR, unfortunately, I need to rewrite the whole thing if I want to provide a good exapmle for dynamic stores. I created https://github.com/orgs/vuejs/projects/21/views/1?query=sort%3Aupdated-desc+is%3Aopen&pane=issue&itemId=136739859 to track

PR #1329, opened 2022-06-01
docs: nuxt mention

dosstx:
Official Nuxt3 docs state that the `buildModules` array is @deprecated — This is no longer needed in Nuxt 3 and Nuxt Bridge; all modules should be added to modules instead.

### DIFF
--- a/packages/nuxt/README.md
+++ b/packages/nuxt/README.md
@@ -10,7 +10,7 @@ npm i @pinia/nuxt
 
 ## Usage
 
-Add to `buildModules` in `nuxt.config.js`:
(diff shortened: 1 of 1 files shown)


packages/nuxt/README.md
```diff
@@ -13,9 +13,14 @@ npm i @pinia/nuxt
 Add to `buildModules` in `nuxt.config.js`:
```
posva:
```suggestion
Add to `modules` (Nuxt 3) or `buildModules` (Nuxt 2) in `nuxt.config.js`:
```

PR #1356, opened 2022-06-10
feat: warn when a getter conflicts with the state

### LINKED ISSUE #1331: Warn when Getter with same name as State
### Reproduction

https://github.com/TorchlightSoftware/nuxt-bugs/tree/pinia-getter-state-conflict

Create a getter with the same name as a property in the state.

```ts
import {defineStore} from 'pinia'

export const useAuthStore = defineStore('auth', {
  state: () => ({
    user: {fname: 'Donna', lname: 'Renolds'},
  }),

  getters: {
    user: (state) => state.user,
  },

  actions: {
  },
})

```

### Expected behavior

This should produce an error telling the application developer not to do this.

### Actual behavior

When the value is consumed, it overflows the stack:

```
[nuxt] [request error] Maximum call stack size exceeded
  at formatRaw (node:internal/util/inspect:1037:12)  
  at formatValue (node:internal/util/inspect:817:10)  
  at inspect (node:internal/util/inspect:347:10)  
  at formatWithOptionsInternal (node:internal/util/inspect:2167:40)  
  at formatWithOptions (node:internal/util/inspect:2029:10)  
  at console.value (node:internal/console/constructor:324:14)  
  at console.log (node:internal/console/constructor:360:61)  
  at Proxy.me (./.nuxt/dist/server/server.mjs:2956:15)  
  at ReactiveEffect.fn (./node_modules/pinia/dist/pinia.mjs:1159:38)  
  at ReactiveEffect.run (./node_modules/@vue/reactivity/dist/reactivity.cjs.js:168:25)
```

### Additional information

Beginner mistake, not understanding that the getter is unnecessary in this case.  A clear error message would save time and help to more quickly determine the cause.  Thanks!

BlackCrowxyz:
### Description:
This allows Pinia store to warn user if they are using the same getter and state property name in a store.

### Related Issue:
close #1331 

### Testing Notes:
Passed all tests, didn't add any new test

### Side Note:
I think showing the message using `consloe.error` is more understandable than `consloe.warn`. Because in some cases equal names leads to furthur errors.

1. using `consloe.error`:

photo_2022-06-10_12-52-10

2. using `consloe.warn`:
photo_2022-06-10_12-52-12


packages/pinia/src/store.ts
```diff
@@ -149,6 +149,17 @@ function createOptionsStore<
       localState,
       actions,
       Object.keys(getters || {}).reduce((computedGetters, name) => {
+        if (__DEV__) {
+          Object.keys(pinia.state.value[id] || {}).forEach((stateName) => {
```
posva: Let's check if the name exists in `localState`, it will be much faster

packages/pinia/src/store.ts
```diff
@@ -149,6 +149,17 @@ function createOptionsStore<
...
+            if (name === stateName) {
+              console.error(
```
```suggestion
              console.warn(
```

packages/pinia/src/store.ts
```diff
@@ -149,6 +149,17 @@ function createOptionsStore<
...
+                `[🍍]: "state" property name cannot be the same as "getters" property name.\n` +
+                  `\tGetter "${name}" is the same as state "${stateName}" in "${id}" store.`
```
```suggestion
                `[🍍]: A getter cannot have the same name as another state property. Rename one of them. Found with "${name}" in store "${id}".`
```

posva (changes requested):
Can you add a test? There are other tests for warns so it should be easy to copy one.

packages/pinia/__tests__/store.spec.ts
```diff
@@ -366,4 +366,17 @@ describe('Store', () => {
     useMyStore()
     expect(warnTextCheckPlainObject('poInit')).toHaveBeenWarnedTimes(0)
   })
+
+  it('only warns when state name conflicts with getters name', () => {
+    const useStore = defineStore({
+      id: 'main',
```
posva: Pass it as the first argument instead

packages/pinia/src/store.ts
```diff
@@ -150,7 +150,7 @@ function createOptionsStore<
       actions,
       Object.keys(getters || {}).reduce((computedGetters, name) => {
         if (__DEV__) {
-          Object.keys(pinia.state.value[id] || {}).forEach((stateName) => {
+          Object.keys(localState || {}).forEach((stateName) => {
```
posva: No, but just use the in operator instead of forEach 😅

.vscode/settings.json
```diff
@@ -3,7 +3,8 @@
     "editor.formatOnSave": true
   },
   "[typescript]": {
-    "editor.formatOnSave": true
+    "editor.formatOnSave": true,
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
posva: Remove this please

posva (changes requested):

packages/pinia/src/store.ts
```diff
@@ -149,14 +149,10 @@ function createOptionsStore<
...
       Object.keys(getters || {}).reduce((computedGetters, name) => {
-        if (__DEV__) {
-          Object.keys(localState || {}).forEach((stateName) => {
-            if (name === stateName) {
-              console.warn(
-                `[🍍]: A getter cannot have the same name as another state property. Rename one of them. Found with "${name}" in store "${id}".`
-              )
-            }
-          })
+        if (__DEV__ && name in (localState || {})) {
```
posva: do we really need the `|| {}` check?

```suggestion
        if (__DEV__ && name in localState) {
```

posva (approved):

packages/pinia/__tests__/store.spec.ts
BlackCrowxyz: What is the difference between passing store name(id) as the first argument and using it as property?
Other tests weren't using id as the first argument!

PR #1365, opened 2022-06-11
chore: migrate `vitest` from `jest`

webfansplz:
vitest-pinia


posva:
Thanks but I had some runs and there aren't any changes in speed so let's stay on jest here as we don't really benefit from having vite plugins this codebase.

PR #1374, opened 2022-06-15
docs: create a pinia example (#1374) [skip ci]

schalkneethling:
The current example caused a bit of confusion for me:

```js
import { createPinia } from 'pinia'

app.use(createPinia())
```

After some trial and error and reading the errors in devtools I realized that what I needed was the following in `main.js`(Note I am using Vue3 with Vite)

```js
import { createApp } from "vue";
import { createPinia } from "pinia";

import App from "./App.vue";

const pinia = createPinia();
const app = createApp(App);

app.use(pinia);
app.mount("#app");
```


posva:
Thanks!

schalkneethling:
> Thanks!

Thank you for merging and the amazing library!

PR #1384, opened 2022-06-20
docs: fix line highlights

lazzzis:
Per [1](https://github.com/vuejs/pinia/blob/667b81dfd7888bbae562bbd02c3670d4c664a8e2/packages/docs/.vitepress/config.js#L31) and [2](https://github.com/vuejs/vitepress/blob/75ca9e4302c65e3bcc9518f7df928318380f6cf6/src/node/markdown/plugins/highlightLines.ts#L13), replace `{` & `}` with `%{` and `}%` because we customized attrs

Before:
Complain that the language does not exist
Screen Shot 2022-06-20 at 12 31 22 AM
Code is not correctly highlighted:
Screen Shot 2022-06-20 at 12 32 42 AM

After:
Screen Shot 2022-06-20 at 1 00 14 AM
Screen Shot 2022-06-20 at 1 02 28 AM

### DIFF
--- a/packages/docs/.vitepress/theme/code-theme.css
+++ b/packages/docs/.vitepress/theme/code-theme.css
@@ -303,6 +303,11 @@ html.light .token.italic {
   font-style: italic;
 }
 
+/* Fix https://github.com/vuejs/vitepress/issues/63 */
+.highlight-lines {
+  padding: 20px 0 0 0;
+}
+
 .highlight-lines .highlighted {
   background-color: var(--code-highlight-bg-color);
 }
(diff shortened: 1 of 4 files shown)


posva:
Thanks! I think there is something else wrong because it currently looks like this:

Screenshot 2022-06-20 at 10 18 11

lazzzis:
This seems related to vuejs/vitepress#63.  Unless you are fine with bumping vitepress to 1.0.0-alpha, I added some styles to fix that, based on the [new theme styles](https://github.com/vuejs/vitepress/blob/75ca9e4302c65e3bcc9518f7df928318380f6cf6/src/client/theme-default/styles/components/vp-doc.css#L322). Since the font size is larger than the default one, the padding-top value is a little larger accordingly.

posva:
Thanks for this. I went with https://github.com/vuejs/pinia/pull/1387 because we want to change the theme in the near future and I want to avoid adding any changes I will have to remove.

PR #1392, opened 2022-06-24
fix: install devtools plugin only after app is created and installed (#1063)

### LINKED ISSUE #1063: Nuxt 2 - using a store in a plugin outputs an error in console
### Reproduction

https://github.com/guillaumeduliscouet/example-nuxt-2

It is a fork of the official exemple with only this added as a nuxt plugin:

```js
export default async ({ $pinia }) => {
    useUserStore($pinia).logout()
}
```

### Steps to reproduce the behavior

1. Open the app
2. Open the vue devtools
3. Wait 1 minute
4. See the error

### Expected behavior

Using a store in a plugin should be possible (right ?) without breaking the devtools

### Actual behavior

Everything seems to work in the devtools, but:
- the message `"user" store installed` does not appear
- after 1 minute we have an error message:

image

Here is what I have if I remove the store usage in the plugin:

image

### Additional information

I imagine that the plugin gets called before the Vue app creation ? Seems like pinia is trying to install it's devtools plugin with `app: null`

popov-a-e:
Closes https://github.com/vuejs/pinia/issues/1063

Turned out in nuxt plugin environment, `pinia._a` is not yet defined which resulted in devtools unable to find app record.

### DIFF
--- a/packages/pinia/src/createPinia.ts
+++ b/packages/pinia/src/createPinia.ts
@@ -18,6 +18,8 @@ export function createPinia(): Pinia {
   let _p: Pinia['_p'] = []
   // plugins added before calling app.use(pinia)
   let toBeInstalled: PiniaPlugin[] = []
+  let afterInitializeCallbacks: ((app: App) => void)[] = []
+  let _a: App | null = null
 
   const pinia: Pinia = markRaw({
     install(app: App) {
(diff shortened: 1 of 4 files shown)


posva:
Thanks for the PR but this completely changes the behavior of plugins. The fix should be possible with minimal changes and without changing how stores are initialized

popov-a-e:
Vue-devtools requires vue instance (app) for a plugin to be properly installed.
This instance might be uninitialized when useStore is called, there is no need for nuxt to reproduce that:
```
const pinia = createPinia()
useFooStore(pinia)
createApp(App).use(pinia)
```
So the only proper solution is to install (at least some) plugins only after app is created and installed to pinia.
I can do it via adding optional argument to `pinia.use` or create `pinia.useWithApp`, or extend `pinia.use` to allow adding plugins post-initialization.
Non of these seem beautiful to me, but I don't see better way without refactoring and with _minimal changes_, what do you think?

popov-a-e:
OK, I ended up applying same logic to devtools plugin only.
Tests checked.

popov-a-e:
@posva Hi Eduardo, is it OK now?

posva (changes requested):
Thanks but we really want to avoid adding any new hook like this. 
Instead, let's move this kind of logic to the devtools: if the `registerPiniaDevtools()` haven't been called yet, we buffer the pending stores and once the devtools are registered, we add the store to the devtools.

PR #791, opened 2021-11-10
feat(nuxt)!: rewrite nuxt module

pi0:
This PR rewrites the Pinia+Nuxt module for best Nuxt 3 and Bridge compatibility 💛 

Testing repo with published package: https://github.com/pi0/nuxt-tests/tree/pinia

Changes:
- Add pinia as a dependency -- Nuxt modules are supposed to be standalone and provide integration out of the box
- Migrate back to kit -- However using kit is not documented yet and might change, it will make stability much better than not using it and without it we have to inline bridge support
- Split vue 2 (bridge) and vue 3 plugins
- Remove deprecated `ctx.pinia` 
- Nuxt 2 without bridge is not supported anymore
- Drop `.$nuxt` support to prefer `useNuxtApp`

TODO:
- [ ] Use resolveModule from kit to respect search path for pinia
- [ ] Detect bridge || nuxt 3 (https://github.com/nuxt/nuxt.js/issues/11869)
- [ ] Fix config types
- [ ] Fix runtime types
- [x] Auto import for pinia

PR #1435, opened 2022-07-08
feat!(nuxt): upgrade nuxt module 

### LINKED ISSUE #1449: Nuxt3 - Cannot set properties of undefined (setting 'pinia')
### Reproduction

https://stackblitz.com/edit/github-1xxy9u

[warn] [nuxt] You are using a plugin with legacy Nuxt 2 format (context, inject) which is likely to be broken. In the future they will be ignored: PiniaNuxtPlugin

### Expected behavior

Works as expected

### Actual behavior

Can not start nuxt3 project

### Additional information

_No response_

pi0:
Rework of #791

- Add pinna as a dependency -- Nuxt modules are supposed to be standalone and provide integration out of the box
- Add simple playground 
- Migrate to [module-builder](https://github.com/nuxt/module-builder) and [nuxt kit](https://v3.nuxtjs.org/guide/going-further/kit)
- Split vue 2 (bridge) and vue 3 plugins
- Remove deprecated `ctx.pinia` 
- Nuxt 2 support requires bridge to be enabled (using plain simple plugin)
- Use `nuxi typecheck` for tpyechecking module
- `vue-demi` is removed for vue2 support since bridge is in place
- Auto imported composables (`definePiniaStore()` and `usePinia()` exposed)

Breaking changes:
- Nuxt 2 users need to enable [nuxt bridge](https://github.com/nuxt/bridge) in order to use latest version of the module.
- Only `$pinia` injection is available which can be accessed using `usePinia()` (or `useNuxtApp().$pinia`)
- `$nuxt` from store can be accessed using `useNuxtApp()` composable

To be tested:
- [x] Nuxt 3 on an external project
- [x] Nuxt 2 (w, w/o bridge) on an external project


packages/nuxt/package.json
```diff
@@ -24,35 +24,34 @@
...
+    "dev": "nuxi dev playground",
+    "dev:build": "nuxi build playground",
+    "dev:prepare": "nuxt-module-build --stub && nuxi prepare playground",
+    "test:types": "nuxi typecheck",
     "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s --commit-path . -l @pinia/nuxt -r 1"
   },
   "dependencies": {
-    "vue-demi": "*"
+    "@nuxt/kit": "^3.0.0-rc.4",
+    "pinia": ">=2.0.14"
```
posva, line 49: Is this really a good idea? Then, it will be more difficult to upgrade pinia for users. Given how package managers error when you don't install the peer dependencies it feels like having a peer dep is a better direction.

packages/nuxt/src/runtime/composables.ts
```diff
@@ -0,0 +1,6 @@
+import { useNuxtApp } from '#imports'
+import { defineStore } from 'pinia'
+
+export const usePinia = () => useNuxtApp().$pinia
```
posva, line 4: do we really need this? This sohuld be the same as `getActivePinia()`

packages/nuxt/src/module.ts
```diff
@@ -1,70 +1,59 @@
...
+      addPlugin(resolver.resolve('./runtime/plugin.vue2'))
+    } else {
+      addPlugin(resolver.resolve('./runtime/plugin.vue3'))
+    }
+
+    // Add auto imports
+    const composables = resolver.resolve('./runtime/composables')
+    addAutoImport([
+      { from: composables, name: 'usePinia' },
+      { from: composables, name: 'definePiniaStore' },
```
posva: Why the rename?

packages/nuxt/package.json
pi0: Nuxt itself has a direct dependency on `vue`, `vue-router` and other libs and Nuxt modules do the same. `<pm> upgrade @pinia/nuxt` updates all of its sub dependencies within supported semver range (here is 2.x) (therefore the latest version of pinia will be installed). This gives one package installation DX. It also helps in some rare cases that you directly control working dependency version without asking users to manually do.

Regardless I'm fine with either direction you like :)

packages/nuxt/src/module.ts
```diff
@@ -1,70 +1,59 @@
...
-
-  this.addPlugin({ src: require.resolve('./plugin.mjs') })
-
-  // transpile pinia for nuxt 2 and nuxt bridge
-  // if (isVue2 && !nuxt.options.build.transpile.includes('pinia')) {
-  //   nuxt.options.build.transpile.push('pinia')
-  // }
-}
-
-declare module '@nuxt/types' {
```
posva, line 47: if we remove these we won't have access to `$nuxt.$pinia`. Unless nuxt is somehow adding the types magically ✨

packages/nuxt/src/runtime/composables.ts
pi0: Hmm then i guess not if we already have a composable to access it. Does it return same reference of `$pinia`? (then probably plugin injection is not necessary too). 

One small benefit of nuxt composables vs vue composables is that they also work outside of vue setup lifecycle with better accessibility support.

packages/nuxt/src/module.ts
pi0: As of [module author guide](https://v3.nuxtjs.org/guide/going-further/modules#always-prefix-exposed-interfaces), having prefix avoids possible conflict between two modules exposing same name for auto imports.

packages/nuxt/src/module.ts
pi0: Indeed it does infer when returning `{ provide }` from plugins :)

posva: but what about Nuxt 2?

packages/nuxt/src/runtime/plugin.vue2.ts
```diff
@@ -10,6 +10,13 @@ export default (context: any, inject: any) => {
   context.app.pinia = pinia
   setActivePinia(pinia)
 
+  // add access to `$nuxt`
+  // @ts-expect-error: _p is internal
+  pinia._p.push(({ store }) => {
+    // make it non enumerable so it avoids any serialization and devtools
+    Object.defineProperty(store, '$nuxt', { value: context })
```
pi0, line 17: Nuxt context (app) can be accessed with useNuxtApp() global. Any specific usecase for $nuxt as a plugin?

packages/nuxt/src/runtime/plugin.vue2.ts
```diff
@@ -21,3 +28,13 @@ export default (context: any, inject: any) => {
   // Inject $pinia
   inject('pinia', pinia)
 }
+
+declare module 'pinia' {
+  export interface PiniaCustomProperties {
+    /**
+     * Nuxt context.
```
pi0: It is NuxtApp from @nuxt/schema if you need to manually extend...

posva:
I force pushed because due to conflicts in pnpm-lock

packages/nuxt/src/module.ts
pi0: Having bridge enabled, same DX should be provided (ee need to use plugin wrapper for vue2 pligin in this case)

posva:
The types never worked on my end. It's as if vue-tsc didn't pick up the correct tsconfig so I removed them from CI

packages/nuxt/src/module.ts
posva: hmm, I think in this case, pinia can keep the defineStore and others will use a prefix

packages/nuxt/src/runtime/plugin.vue2.ts
posva: People use to access `$fetch` and other properties. Should they not? I was quite requested back when I added it

posva:
Do you also get this one:

Screenshot 2022-07-08 at 17 38 27

It didn't appear before...

posva:
So I tested and this only works with Nuxt 3 😓 
It fails with Nuxt 2 and Nuxt bridge with different errors

- In Nuxt 2 it seems it's importing two different versions of Pinia but I haven't been able to debug further
- In Nuxt Bridge, it's yielding 

```
 WARN  Disabling duplicate auto import 'undefined' (imported from '/Users/posva/piniajs/example-nuxt-bridge/node_modules/@pinia/nuxt/dist/runtime/composables')
```

and 

```
Cannot read properties of undefined (reading 'install')
  at Function.Vue.use (file://./.nuxt/dist/server/server.mjs:7840:23)
  at $id_6aebe68d (file://./.nuxt/dist/server/server.mjs:22659:5)
  at async __instantiateModule__ (file://./.nuxt/dist/server/server.mjs:22861:3)
```

packages/nuxt/src/runtime/plugin.vue2.ts
pi0: For Nuxt 2 yes made sense, but in Nuxt 3 (and with bridge), we have global `$fetch` and composables replacing the old `$nuxt` API.

packages/nuxt/src/module.ts
pi0: [Global store RFC](https://github.com/nuxt/framework/discussions/571) is still open. If we add it as built-in and even if based on pinna for implementation, module and built-in might conflict and introduce slight breaking change. If you prefer to advocate `defineStore`, we can update example to add explicit `import { defineStore } from 'pinia'` :)

packages/nuxt/src/runtime/plugin.vue2.ts
posva: Should I mark it as deprecated then? Is there a page that makes sense to point in the jsdoc?

pi0:
Thanks for the review @posva. I will rebase and check last changes on bridge.

packages/nuxt/src/runtime/plugin.vue2.ts
pi0: It is (currently briefly) documented here: https://v3.nuxtjs.org/guide/going-further/nuxt-app

Since this PR is breaking change, I would drop it. `$nuxt` interface in Nuxt 2 is not same as `NuxtApp`. It might be confusing to keep (still your choice)

packages/nuxt/src/module.ts
```diff
@@ -1,70 +1,62 @@
...
+
+    // Make sure we use the mjs build for pinia
+    nuxt.options.alias.pinia = 'pinia/dist/pinia.mjs'
+
+    // Add runtime plugin
+    if (isNuxt2()) {
+      addPlugin(resolver.resolve('./runtime/plugin.vue2'))
+    } else {
+      addPlugin(resolver.resolve('./runtime/plugin.vue3'))
+    }
```
Aareksio, line 70: > Note: By default plugin is prepended to the plugins array. You can use second argument to append (push) instead.

This means, that by default, if we precise an array of modules like so:
```
modules: ['@pinia/nuxt', 'my-module']
```

We'll receive `getActivePinia was called with no active Pinia` error, as any plugin added by `my-module` with default options is going to be executed **before** pinia plugin can call `setActivePinia()`.

See reproduction: https://stackblitz.com/edit/github-ba6rwp?file=nuxt.config.ts

---

Raised as issue in nuxt: https://github.com/nuxt/framework/issues/5835

pi0: You shouldn't depend on module order (please see my comment: https://github.com/nuxt/framework/issues/5835#issuecomment-1181679108). We might improve with a new hook to allow extending pinia via other module plugins.

packages/nuxt/src/module.ts
pi0: (make `definePiniaStore` opt-in since is not standard)

packages/nuxt/src/runtime/plugin.vue3.ts
```diff
@@ -0,0 +1,37 @@
+import { createPinia, setActivePinia } from 'pinia'
+import { defineNuxtPlugin } from '#imports'
+
+export default defineNuxtPlugin((nuxtApp) => {
+  const pinia = createPinia()
+  nuxtApp.vueApp.use(pinia)
+  setActivePinia(pinia)
+
+  // @ts-expect-error: _p is internal
+  pinia._p.push(({ store }) => {
```
pi0: (TODO: Remove from nuxt 3)

packages/nuxt/src/runtime/plugin.vue2.ts
pi0: (to be deprecated)

PR #1441, opened 2022-07-10
docs: remove unecessary religious references that can offend some people

minenwerfer:
Those are really unneeded.


posva:
I actually wanted to change this a while ago but not for the same reason. It's more of a reference to Hereditary, the horror movie, and the demon lore in 18th-century books, I think it's not relatable enough so I added a JoJo reference and called it a day

PR #1444, opened 2022-07-12
fix(types): setupStore getter types (#1430)

### LINKED ISSUE #1430: Typescript type of getter in composed store is wrong
### Reproduction

https://github.com/DerZade/pinia-bug-report

Open the `App.vue` in VS Code or any other editor, that supports typescript type-checking IntelliSense. `counter.nPlusX` is inferred as type `number`, although it actually is a `(x: number) => number`.

### Expected behavior

A computed property of a composed store, that returns a function (to achieve [the effect of a getter, to which I can pass arguments](https://pinia.vuejs.org/core-concepts/getters.html#passing-arguments-to-getters)) should have a correctly inferred Typescript type.

### Actual behavior

A computed property of a composed store, that returns a function, has a inferred type of the return type of said function.

### Additional information

_No response_

Derhelios66:
I have an Idea to fix the following Issue. :smile:
Fix https://github.com/vuejs/pinia/issues/1430

### DIFF
--- a/packages/pinia/src/types.ts
+++ b/packages/pinia/src/types.ts
@@ -601,7 +601,7 @@ export type _ExtractActionsFromSetupStore<SS> = SS extends undefined | void
 export type _ExtractGettersFromSetupStore<SS> = SS extends undefined | void
   ? {}
   : _ExtractGettersFromSetupStore_Keys<SS> extends keyof SS
-  ? _UnwrapAll<Pick<SS, _ExtractGettersFromSetupStore_Keys<SS>>>
+  ? Pick<SS, _ExtractGettersFromSetupStore_Keys<SS>>
   : never
 
 /**
--- a/packages/pinia/test-dts/storeSetup.test-d.ts
+++ b/packages/pinia/test-dts/storeSetup.test-d.ts
@@ -11,13 +11,17 @@ const useSetupStore = defineStore('name', () => {
     },
   })
 
+  const multiply = computed(() => (n: number) => {
+    return count.value * n;
+  });
+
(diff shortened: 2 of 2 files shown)


packages/pinia/src/types.ts
```diff
@@ -601,7 +601,7 @@ export type _ExtractActionsFromSetupStore<SS> = SS extends undefined | void
 export type _ExtractGettersFromSetupStore<SS> = SS extends undefined | void
   ? {}
   : _ExtractGettersFromSetupStore_Keys<SS> extends keyof SS
-  ? _UnwrapAll<Pick<SS, _ExtractGettersFromSetupStore_Keys<SS>>>
```
posva, line 604: I need to check if removing this also breaks the <kbd>ctrl</kbd> + click in IDEs

Derhelios66:
Dont worry i've just rebased the branch 😄

bodograumann:
I had to update my mock wrapper after updating to 2.0.17 by adding `UnwrapRef` to the getters:
```typescript
type Mocked<T> = { [P in keyof T]: ReturnType<typeof jest.fn> };
function mocked<Id extends string, S, G, A>(store: Store<Id, S, G, A>): S & UnwrapRef<G> & Mocked<A> {
  /* eslint-disable-next-line @typescript-eslint/ban-ts-comment -- the new type is due to `app.use(createTestingPinia())` */
  //@ts-ignore
  return store;
}
```

`G` was previously inferred as having plain properties, but now it somehow contains `ComputedRef`s.
I hope that this info might help others.

Issue #1394, opened 2022-06-26
Warn if two stores are defined with the same name


Soviut:
### What problem is this solving

I have a store. I copied it to make a second store but forgot to change the name of the second store.

This resulted in very subtle bugs where store state that should never be able to be `undefined` was appearing as `undefined`.

It should be noted that these stores are being intentionally used in an un-strict manner where the state is being mutated directly.

**store/invoices.ts**
```typescript
export const useStore = defineStore('invoices', () => {
```
**store/templates.ts**
```typescript
export const useStore = defineStore('invoices', () => {
```

### Proposed solution

Warn if multiple stores share the same name.

### Describe alternatives you've considered

Using the Vue Dev Tools, I was able to see that only one store existed and it sometimes had the values of the first store and sometimes values from the second store.

jamiecarter7:
I had a simlar issue, where I had created a state variable and a function with the same name.

I was getting some 'computed' error on my nuxt project until I did some digging

themeler:
Since Vuex there is one thing in my mind that could resolve the issue with possible name duplicates. Personally I have one big enum for a whole project (with store names) but it feels like a workaround.

Maybe instead of using `string` as a store name, there could be at least an option to use `symbol`. It's description could serve as displayed store name in dev tools. That would give unique id for every store, even if the name is the same. It could still warn about duplicated names (symbol descriptions) but without failing functionally.

PR #1452, opened 2022-07-13
feat: add warning message if store id already exists (#1394)

zavsievich:
Add a toast message if store with current id already exists.

### DIFF
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -904,7 +904,7 @@ export function defineStore(
       }
     } else {
       toastMessage(
-        `Store with id: ${id} already exists. Stores shoud have unique id.`,
+        `Store with id: ${id} already exists. Stores should have unique id.`,
         'warn'
       )
     }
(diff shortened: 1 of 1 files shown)


posva:
Thanks but this doesn't do what you think it does 😓 , it doesn't add the feature you linked. Did you try it?

zavsievich:
> Thanks but this doesn't do what you think it does 😓 

Could you please explain how to fix this?

packages/pinia/src/store.ts
```diff
@@ -900,6 +902,11 @@ export function defineStore(
         // @ts-expect-error: not the right inferred type
         useStore._pinia = pinia
       }
+    } else {
+      toastMessage(
+        `Store with id: ${id} already exists. Stores shoud have unique id.`,
```
BlackCrowxyz:
```suggestion
        `Store with id: ${id} already exists. Stores should have unique id.`,
```

BlackCrowxyz:
> > Thanks but this doesn't do what you think it does 😓
> 
> Could you please explain how to fix this?

When changing routes in playground, you can see unnecessary warning messages. Pinia registers stores with `id` when creating them, reusing stores in other places mean they were registered before with the `id` so the warning will come up!

OXXXRA (approved):

OXXXRA (approved):

posva:
Thanks but this does not implement the linked feature, it warns whenever the store is reused (which we don't want ofc).

We don't know if we will implement the feature yet but if you want to give it another shot make sure to:

- include tests
- make the warning dev only (just a warn, not the toast)

PR #1466, opened 2022-07-22
Add documentation about optional state properties

jume-dev:
I only found out about it by searching through the issues, so I guess it's useful to have something like this in the docs. I am not 100% confident in the wording but I wanted to make a first shot to get a discussion going.


posva:
Thanks, I ended up a longer explanation

PR #1467, opened 2022-07-22
docs: clarify composing-stores.md, remove duplicate example (#1467) [skip ci]

bashunaimiroy:
We had a code example with a getter and an action, then one with the getter, then one with the action.
I removed the first one.

I also clarified right up at the front that sharing state/methods between stores is supported in Pinia, since I believe a lot of people will find this doc by searching that exact question.

I also removed a mention of 'nesting' because we use that specific term elsewhere to refer to a file/folder structure, and it doesn't seem to me to be semantically useful - it's just a store using a store, I don't think there's any hierarchy implied.

### DIFF
--- a/packages/docs/cookbook/composing-stores.md
+++ b/packages/docs/cookbook/composing-stores.md
@@ -42,7 +42,7 @@ const useY = defineStore('y', () => {
 
 ## Using a store within another store
 
-Note that if one store uses another store, you can directly import and call the `use${storeName}Store` method, and access methods on the store object, just like you would access them from within a Vue component.
+Note that if one store uses another store, you can directly import and call the `use${storeName}Store` method, then access methods on the returned store object, just like you would access them from within a Vue component.
 
 You can call `useOtherStore()` inside any getter or action:
 
--- a/packages/docs/cookbook/composing-stores.md
+++ b/packages/docs/cookbook/composing-stores.md
@@ -1,6 +1,6 @@
 # Composing Stores
 
-Composing stores is about having stores that use each other and there is one rule to follow:
+Composing stores is about having stores that use each other's state / methods, and this is supported in Pinia. There is one rule to follow:
 
 If **two or more stores use each other**, they cannot create an infinite loop through _getters_ or _actions_. They cannot **both** directly read each other state in their setup function:
 
@@ -40,34 +40,12 @@ const useY = defineStore('y', () => {
 })
 ```
 
-## Nested stores
+## Using a store within another store
 
-Note that if one store uses another store, **there is no need to create a new store in a separate file**, you can directly import it. Think of it as nesting.
+Note that if one store uses another store, you can directly import and call the `use${storeName}Store` method, and access methods on the store object, just like you would access them from within a Vue component.
 
-You can call `useOtherStore()` at the top of any getter or action:
+You can call `useOtherStore()` inside any getter or action:
 
-```js
-import { useUserStore } from './user'
-
-export const cartStore = defineStore('cart', {
-  getters: {
-    // ... other getters
-    summary(state) {
-      const user = useUserStore()
-
-      return `Hi ${user.name}, you have ${state.list.length} items in your cart. It costs ${state.price}.`
-    },
-  },
-
-  actions: {
-    purchase() {
-      const user = useUserStore()
-
-      return apiPurchase(user.id, this.list)
-    },
-  },
-})
-```
 
 ## Shared Getters
 
--- a/packages/docs/cookbook/composing-stores.md
+++ b/packages/docs/cookbook/composing-stores.md
@@ -1,6 +1,6 @@
 # Composing Stores
 
-Composing stores is about having stores that use each other's state / methods, and this is supported in Pinia. There is one rule to follow:
+Composing stores is about having stores that use each other, and this is supported in Pinia. There is one rule to follow:
 
 If **two or more stores use each other**, they cannot create an infinite loop through _getters_ or _actions_. They cannot **both** directly read each other state in their setup function:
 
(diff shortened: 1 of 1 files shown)


packages/docs/cookbook/composing-stores.md
```diff
@@ -40,34 +40,12 @@ const useY = defineStore('y', () => {
 })
 ```
 
-## Nested stores
+## Using a store within another store
 
-Note that if one store uses another store, **there is no need to create a new store in a separate file**, you can directly import it. Think of it as nesting.
+Note that if one store uses another store, you can directly import and call the `use${storeName}Store` method, and access methods on the store object, just like you would access them from within a Vue component.
```
bashunaimiroy:
```suggestion
Note that if one store uses another store, you can directly import and call the `use${storeName}Store` method, then access methods on the returned store object, just like you would access them from within a Vue component.
```

packages/docs/cookbook/composing-stores.md
```diff
@@ -40,34 +40,12 @@ const useY = defineStore('y', () => {
 })
 ```
 
-## Nested stores
+## Using a store within another store
 
-Note that if one store uses another store, **there is no need to create a new store in a separate file**, you can directly import it. Think of it as nesting.
```
bashunaimiroy, line 45: This seemed likely to mislead beginners- you do indeed need to define a store in a separate file

packages/docs/cookbook/composing-stores.md
```diff
@@ -1,6 +1,6 @@
 # Composing Stores
 
-Composing stores is about having stores that use each other and there is one rule to follow:
+Composing stores is about having stores that use each other's state / methods, and this is supported in Pinia. There is one rule to follow:
```
posva: let's keep the old version of just saying _use each other_

packages/docs/cookbook/composing-stores.md
```diff
@@ -40,34 +40,12 @@ const useY = defineStore('y', () => {
 })
 ```
 
-## Nested stores
+## Using a store within another store
```
posva: let's keep the old version

packages/docs/cookbook/composing-stores.md
```diff
@@ -40,34 +40,12 @@ const useY = defineStore('y', () => {
...
 
-Note that if one store uses another store, **there is no need to create a new store in a separate file**, you can directly import it. Think of it as nesting.
+Note that if one store uses another store, you can directly import and call the `use${storeName}Store` method, then access methods on the returned store object, just like you would access them from within a Vue component.
 
-You can call `useOtherStore()` at the top of any getter or action:
+You can call `useOtherStore()` inside any getter or action:
```
posva: Let's remove this line

packages/docs/cookbook/composing-stores.md
posva: indeed, this isn't clear. It was referring to creating a 3rd file for an extra store when we already have 2 of them

packages/docs/cookbook/composing-stores.md
```diff
@@ -40,34 +40,12 @@ const useY = defineStore('y', () => {
 })
 ```
 
-## Nested stores
+## Using a store within another store
 
-Note that if one store uses another store, **there is no need to create a new store in a separate file**, you can directly import it. Think of it as nesting.
+Note that if one store uses another store, you can directly import and call the `use${storeName}Store` method, then access methods on the returned store object, just like you would access them from within a Vue component.
```
posva:
```suggestion
Note that if one store uses another store, you can directly import and call the `useStore()` function within _actions_ and _getters_. Then you can interact with the store just like you would from within a Vue component. See [Shared Getters](#shared-getters) and [Shared Actions](#shared-actions).
```

packages/docs/cookbook/composing-stores.md
```diff
@@ -40,34 +40,12 @@ const useY = defineStore('y', () => {
...
 
-You can call `useOtherStore()` at the top of any getter or action:
+You can call `useOtherStore()` inside any getter or action:
 
-```js
-import { useUserStore } from './user'
-
-export const cartStore = defineStore('cart', {
```
posva, line 52: let's keep this example but transform it into the setup store syntax:

```ts
defineStore('cart', () => {
  // ...
})
```

posva (changes requested):
Thanks!

packages/docs/cookbook/composing-stores.md
bashunaimiroy: Hi @posva, thanks for feedback. I've made the other changes but I'm confused by this one, and it might just be my lack of familiarity with the code.

1) I'm not sure how to write this example with the syntax you're asking. I've only found out now that `defineStore` accepts a function rather than an Object as its second argument- Pinia docs mostly demonstrate using an Object, and my org only uses this approach. [It seems like this confusion is shared](https://www.reddit.com/r/vuejs/comments/snpo86/comment/hw6sv1j/?utm_source=reddit&utm_medium=web2x&context=3), and this document is perhaps the only part of the docs that demonstrates passing a `storeSetup` function, in its infinite loop example above. There's probably already a push to document this more, but maybe we could start with other parts of the docs like [Defining A Store](https://pinia.vuejs.org/core-concepts/#defining-a-store)? I can make a PR if that's helpful
2) I removed this example because it seemed redundant to demonstrate the nested stores twice- we're already demonstrating it in the next two sections. Are you suggesting we keep it just to demonstrate the storeSetup syntax?

bashunaimiroy: update: [I've made a PR](https://github.com/vuejs/pinia/pull/1475) to document the setup store syntax elsewhere in the docs. 

I see that "What is Pinia?" also [demonstrates this](https://pinia.vuejs.org/introduction.html#basic-example:~:text=You%20can%20even%20use%20a%20function%20(similar%20to%20a%20component%20setup())%20to%20define%20a%20Store%20for%20more%20advanced%20use%20cases%3A), but I suspect that a lot of developers won't look there to figure out how `defineStore` works

posva: Cool, thanks!

I will add the example after merging

PR #1471, opened 2022-07-23
docs: The description of $state does not match the pinia source code implementation

coderwhy:
The implementation of $state in the source code is not to replace the whole state of a store, but to merge it.
It does not match the results in practical applications. It is more appropriate to modify it to merge the incoming objects here.
image

### DIFF
--- a/packages/docs/core-concepts/state.md
+++ b/packages/docs/core-concepts/state.md
@@ -147,9 +147,9 @@ cartStore.$patch((state) => {
 
 The main difference here is that `$patch()` allows you to group multiple changes into one single entry in the devtools. Note **both, direct changes to `state` and `$patch()` appear in the devtools** and can be time traveled (not yet in Vue 3).
 
-## Replacing the `state`
+## Merging the `state`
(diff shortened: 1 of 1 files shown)


packages/docs/core-concepts/state.md
```diff
@@ -147,9 +147,9 @@ cartStore.$patch((state) => {
 
 The main difference here is that `$patch()` allows you to group multiple changes into one single entry in the devtools. Note **both, direct changes to `state` and `$patch()` appear in the devtools** and can be time traveled (not yet in Vue 3).
 
-## Replacing the `state`
+## Merging the `state`
```
codersjj, line 150: Looks good to me. But the second half of this section does mention that the `state` can be replaced with `pinia.state.value = {}`, so, maybe this paragraph should be titled "Merging/Replacing the `state`"?

posva:
Thanks, merged locally. Remember to check the checkbox to let maintainers push changes to your branch, it's really helpful 🙂

PR #1520, opened 2022-08-04
docs: Fix incorrect hash anchor

niceplugin:
# Remove prefix underbar in hash anchor.

Some header elements in Type Aliases or Interfaces contain an underscore as a prefix.
However, when replaced with an element's ID, the value does not contain an underscore as a prefix.
However, there are many hash anchors with an underscore as a prefix within the document.
Fixed these to not include prefix underscores.
`#_name` → `#name`

# ⚠️Need to check missing files!
The `ModuleOptions.md` file for the pinia_nuxt interface does not exist.
However, on the website everyone will still be directed to that link and everyone will be able to read the document.
Need to determine if the link is incorrect or if deleting the ModuleOptions.md file is a mistake.

### DIFF
--- a/packages/docs/api/interfaces/pinia.DefineStoreOptions.md
+++ b/packages/docs/api/interfaces/pinia.DefineStoreOptions.md
@@ -36,7 +36,7 @@ augment stores with the plugin API.
 
 ### actions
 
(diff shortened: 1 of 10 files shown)


niceplugin:
On the PR preview page, the anchors have not changed.
❗But in my local test the changed anchor works fine.
There is no problem with the source, but there is a problem with the preview, so you may need to close the PR after checking.

posva:
mmh, the thing is these files are generated with typedoc so changing them won't work

niceplugin:
> mmh, the thing is these files are generated with typedoc so changing them won't work

I don't know the typedoc, so I'll close this PR.
I haven't been of any help, but I hope this solves the problem.
thanks.

posva:
Thanks to you! I think this is a bug in vitepress where titles that start with `_` like `_GettersTree` gets the `_` trimmed

PR #1527, opened 2022-08-05
Update Code

zhao-li:
I was trying to get a basic test working by following this bit of code:
```
import { createTestingPinia } from '@pinia/testing'

const wrapper = mount(Counter, {
  global: {
    plugins: [createTestingPinia()],
  },
})
```

Unfortunately, it didn't work, but what ended up working was this bit of code (from this thread: https://github.com/vuejs/pinia/discussions/1096#discussioncomment-2540811):
```
import { describe, it, expect, vi } from 'vitest'
import { createTestingPinia } from '@pinia/testing'

const wrapper = mount(Counter, {
  global: {
    plugins: [
      createTestingPinia({
        createSpy: vi.fn,
      }),
    ],
  },
})
```

### DIFF
--- a/packages/docs/cookbook/testing.md
+++ b/packages/docs/cookbook/testing.md
@@ -82,12 +82,17 @@ npm i -D @pinia/testing
 And make sure to create a testing pinia in your tests when mounting a component:
 
 ```js
+import { describe, it, expect, vi } from 'vitest'
 import { mount } from '@vue/test-utils'
 import { createTestingPinia } from '@pinia/testing'
 
 const wrapper = mount(Counter, {
   global: {
-    plugins: [createTestingPinia()],
+    plugins: [
+      createTestingPinia({
+        createSpy: vi.fn,
+      }),
+    ],
   },
 })
 


posva:
Thanks but this is mentioned below

zhao-li:
@posva thanks for highlighting the section further down in the docs.  I think this is what you are referring to?:
> Specifying the createSpy function
When using Jest, or vitest with globals: true, createTestingPinia automatically stubs actions using the spy function based on the existing test framework (jest.fn or vitest.fn). If you are using a different framework, you'll need to provide a [createSpy](https://deploy-preview-1527--pinia-official.netlify.app/api/interfaces/pinia_testing.TestingOptions.html#createspy) option:
>
> ```
> import sinon from 'sinon'
> 
> createTestingPinia({
>   createSpy: sinon.spy, // use sinon's spy to wrap actions
> })

Would you accept adding a note about this error message:
> Error: [@pinia/testing]: You must configure the `createSpy` option.

to that section so it makes it a bit more obvious for newbies like myself?

If so, I can create a new pull request for that section.

If not, then all good.  Thank you for pointing things out to me.

Have a Terrific Tuesday 🌈

PR #1551, opened 2022-08-16
feat(nuxt): automatically add pinia types

danielroe:
This PR automatically adds the `@pinia/nuxt` types for Nuxt Bridge / Nuxt 3, and moves this section in the docs to the legacy install instructions.


posva (approved):
Looks good! If anything we can add an extra mention about the TypeScript part being for `@pinia/nuxt` < 0.3.0 (I think that's the first version that is non compatible with Nuxt 2

PR #1553, opened 2022-08-16
docs: fix getter typo

JeraldVin:
seems like a typo in the getter name(line no. 227). **doubleCount** should be **doubleCounter** as given in the example(line no. 187)

line no. 225 - 227
```
// gives access to this.doubleCounter inside the component
// same as reading from store.doubleCounter
...mapState(useCounterStore, ['doubleCount'])
```

line no. 187
```
doubleCounter(state) {
    return state.counter * 2
}
```


posva (approved):
Thanks!

PR #1554, opened 2022-08-16
docs: fix typo buildModules to modules for nuxt3

sathishvm:
Need to change the prop name buildModules to modules for nuxt3 example

### DIFF
--- a/packages/docs/ssr/nuxt.md
+++ b/packages/docs/ssr/nuxt.md
@@ -10,7 +10,7 @@ yarn add @pinia/nuxt
 npm install @pinia/nuxt
 ```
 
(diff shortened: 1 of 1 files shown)


packages/docs/ssr/nuxt.md
```diff
@@ -47,7 +47,7 @@ By default `@pinia/nuxt` exposes one single auto import: `usePinia()`, which is
 // nuxt.config.js
 export default {
```
posva: can you add the defineNuxtConfig as above?

posva (changes requested):
Thanks!

posva (changes requested):
Thanks!

packages/docs/ssr/nuxt.md
sathishvm: sure, will add it

PR #1559, opened 2022-08-17
test: migrate to vitest

danielroe:
This migrates the test suite to `vitest` preparatory to adding some tests for `@pinia/nuxt`, which wasn't playing nicely with jest.

Previous: https://github.com/vuejs/pinia/pull/1365


packages/pinia/__tests__/ssr.spec.ts
```diff
@@ -1,6 +1,7 @@
 /**
  * @jest-environment node
```
posva: should this be changed too?

vite.config.ts
```diff
@@ -0,0 +1,42 @@
...
+        '**/src/index.ts',
+        '**/*.d.ts',
+        '**/src/devtools',
+        '**/src/hmr.ts',
+        '**/src/deprecated.ts',
+        '**/src/vue2-plugin.ts',
+      ],
+    },
+    environment: 'happy-dom',
+    globals: true, // Specifically to make createTestingPinia happy
```
posva, line 40: what do you mean? 😮

danielroe: by default vitest doesn't expose global objects as jest does, so `vi` is not defined by default. We could either update the createTestingPinia tests to pass vi.fn. Or we can use this globals option.

packages/pinia/__tests__/ssr.spec.ts
danielroe: Good point. Vitest does respect it, but we might as well!

vite.config.ts
posva: as long as the type of the variable `createSpy()` isn't `any` in the source code 😆

posva (approved):

PR #1561, opened 2022-08-17
build(pinia): remove unused `@vue/devtools-api` require from `pinia.prod.cjs` (#1561) [skip ci]

pi0:
This PR removed unused `require('@vue/devtools-api')` statement in `pinia.prod.cjs`. Probably there is a better workaround and it is a bug in rollup that does not remove external dependency **after treeshake**.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -119,7 +119,11 @@ function createConfig(buildName, output, plugins = []) {
   hasTSChecked = true
 
   const external = ['vue-demi', 'vue', '@vue/composition-api']
-  if (!isGlobalBuild && !(isProductionBuild && isNodeBuild /* pinia.prod.cjs */)) {
+  if (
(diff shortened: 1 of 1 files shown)


rollup.config.js
```diff
@@ -119,7 +119,7 @@ function createConfig(buildName, output, plugins = []) {
   hasTSChecked = true
 
   const external = ['vue-demi', 'vue', '@vue/composition-api']
-  if (!isGlobalBuild) {
+  if (!isGlobalBuild && !(isProductionBuild && isNodeBuild /* pinia.prod.cjs */)) {
```
posva:
```suggestion
  if (
    !isGlobalBuild &&
    // pinia.prod.cjs should not require `@vue/devtools-api` (like Vue)
    !(isProductionBuild && isNodeBuild)
  ) {
```

PR #1562, opened 2022-08-17
test(nuxt): add e2e test

danielroe:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,7 @@ jobs:
       - run: pnpm install
       - run: pnpm run lint
       - run: pnpm run test:types
+      - run: pnpm run -r dev:prepare
       - run: pnpm run test:vitest
       - run: pnpm run build
       - run: pnpm run build:dts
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -27,6 +27,7 @@ jobs:
       - run: pnpm install
       - run: pnpm run lint
       - run: pnpm run test:types
+      - run: pnpm run -r dev:prepare
```
posva, line 30: should we add this to the root package script `test` as well?

.github/workflows/test.yml
```diff
@@ -27,7 +27,6 @@ jobs:
       - run: pnpm install
       - run: pnpm run lint
       - run: pnpm run test:types
-      - run: pnpm run -r dev:prepare
```
posva: I think this was right, it was more about also adding it to the root test script 😅

danielroe: yep! should be back 😁

posva (approved):

PR #1564, opened 2022-08-18
feat: warn when create store with exsist id

### LINKED ISSUE #1394: Warn if two stores are defined with the same name

OXXXRA:
closes https://github.com/vuejs/pinia/issues/1394

### DIFF
--- a/packages/pinia/__tests__/store.spec.ts
+++ b/packages/pinia/__tests__/store.spec.ts
@@ -379,7 +379,7 @@ describe('Store', () => {
       `[🍍]: A getter cannot have the same name as another state property. Rename one of them. Found with "anyName" in store "main".`
     ).toHaveBeenWarnedTimes(1)
   })
-  it('warns when create store with exsist id', async () => {
+  it('warns when creating store with existing id', async () => {
     const id = 'test id';
     const useFirstStore = defineStore(id, {});
     const useSecondStore = defineStore(id, {});
--- a/packages/pinia/__tests__/store.spec.ts
+++ b/packages/pinia/__tests__/store.spec.ts
@@ -379,4 +379,12 @@ describe('Store', () => {
       `[🍍]: A getter cannot have the same name as another state property. Rename one of them. Found with "anyName" in store "main".`
     ).toHaveBeenWarnedTimes(1)
   })
-})
+  it('warns when create store with exsist id', async () => {
+    const id = 'test id';
+    const useFirstStore = defineStore(id, {});
+    const useSecondStore = defineStore(id, {});
+    useFirstStore();
+    useSecondStore();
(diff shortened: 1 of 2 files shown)


isbik (approved):
Ok

OXXXRA:
@posva, hi, it's ok?

posva:
Thanks but this seem like a copy of #1452 with only one test added... **That PR implementation doesn't work**, if you try it locally you will see how the console is flushed with warnings when running the tests... Also, the warning is [still not dev only](https://github.com/vuejs/pinia/pull/1452#issuecomment-1217889592).

OXXXRA:
@posva, now i added output only for dev, also checked warn output in playground.
Снимок экрана 2022-08-31 в 20 08 33

packages/pinia/__tests__/store.spec.ts
```diff
@@ -379,4 +379,12 @@ describe('Store', () => {
       `[🍍]: A getter cannot have the same name as another state property. Rename one of them. Found with "anyName" in store "main".`
     ).toHaveBeenWarnedTimes(1)
   })
-})
+  it('warns when create store with exsist id', async () => {
```
seanohue: This seems like a typo, should be
```suggestion
  it('warns when creating store with existing id', async () => {
```

packages/pinia/src/store.ts
```diff
@@ -866,6 +866,7 @@ export function defineStore(
     options = idOrOptions
     id = idOrOptions.id
   }
+  // push id into stores array
```
seanohue: This comment is confusing here,  should be on the line before the action its describing, or be removed

OXXXRA: Thank you, i remove useless comment!

packages/pinia/__tests__/store.spec.ts
OXXXRA: Thank you, i change this text!

posva:
Thanks but this still doesn't work and my feedback is getting ignored so I prefer closing this so other contributors feel free to open a new PR

PR #1579, opened 2022-08-21
docs: corrections and improvements

JeraldVin:
1) fix incorrect function name 'doubleCount'
`...mapActions(useCounterStore, { myOwnName: 'doubleCount' }),`
to
`...mapActions(useCounterStore, { myOwnName: 'increment' }),`
2) added note to not use arrow function syntax for defining actions(since it doesn't work)
3) added example using setTimeout to demonstrate reactivity
4) added 'export's, more detail, consistent naming and other improvements

### DIFF
--- a/packages/docs/core-concepts/actions.md
+++ b/packages/docs/core-concepts/actions.md
@@ -8,7 +8,7 @@
 Actions are the equivalent of [methods](https://v3.vuejs.org/guide/data-methods.html#methods) in components. They can be defined with the `actions` property in `defineStore()` and **they are perfect to define business logic**:
 
 ```js
-export const useStore = defineStore('main', {
+export const useCounterStore = defineStore('counter', {
   state: () => ({
     count: 0,
   }),
@@ -59,9 +59,9 @@ Actions are invoked like methods:
 ```js
 export default defineComponent({
   setup() {
-    const main = useMainStore()
+    const store = useCounterStore()
     // call the action as a method of the store
-    main.randomizeCounter()
+    store.randomizeCounter()
 
     return {}
   },
@@ -100,7 +100,7 @@ You can directly call any action as a method of the store:
 ```js
 export default {
   setup() {
-    const store = useStore()
+    const store = useCounterStore()
 
     store.randomizeCounter()
   },
@@ -122,7 +122,7 @@ For the following examples, you can assume the following store was created:
 
 import { defineStore } from 'pinia',
 
-const useCounterStore = defineStore('counter', {
+export const useCounterStore = defineStore('counter', {
   state: () => ({
     count: 0
   }),
@@ -136,7 +136,7 @@ const useCounterStore = defineStore('counter', {
 
 ### With `setup()`
 
-While Composition API is not for everyone, the `setup()` hook can make using Pinia easier to work with in the Options API. No extra map helper functions needed!
+While Composition API is not for everyone, the `setup()` hook can make using Pinia easier to work within the Options API. No extra map helper functions needed!
 
 ```js
 import { useCounterStore } from '../stores/counter'
@@ -218,14 +218,14 @@ const unsubscribe = someStore.$onAction(
 unsubscribe()
 ```
 
-By default, _action subscriptions_ are bound to the component where they are added (if the store is inside a component's `setup()`). Meaning, they will be automatically removed when the component is unmounted. If you want to keep them after the component is unmounted, pass `true` as the second argument to _detach_ the _action subscription_ from the current component:
+By default, _action subscriptions_ are bound to the component where they are added (if the store is inside a component's `setup()`). Meaning, they will be automatically removed when the component is unmounted. If you also want to keep them after the component is unmounted, pass `true` as the second argument to _detach_ the _action subscription_ from the current component:
 
 ```js
 export default {
   setup() {
     const someStore = useSomeStore()
 
-    // this subscription will be kept after the component is unmounted
+    // this subscription will be kept even after the component is unmounted
     someStore.$onAction(callback, true)
 
     // ...
--- a/packages/docs/core-concepts/actions.md
+++ b/packages/docs/core-concepts/actions.md
@@ -170,7 +170,7 @@ export default {
     // same as calling from store.increment()
     ...mapActions(useCounterStore, ['increment'])
     // same as above but registers it as this.myOwnName()
-    ...mapActions(useCounterStore, { myOwnName: 'doubleCount' }),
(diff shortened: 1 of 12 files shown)


packages/docs/core-concepts/index.md
```diff
@@ -27,9 +27,9 @@ Similar to Vue's Options API, we can also pass an Options Object with `state`, `
 
 ```js {2-10}
 export const useCounterStore = defineStore('counter', {
-  state: () => ({ count: 0 }),
+  state: () => ({ count: 0, name: 'Eduardo' }),
```
JeraldVin, line 30: Added the name since we extract it here(line no. 106/110)

`const { name, doubleCount } = store`

packages/docs/core-concepts/index.md
```diff
@@ -50,19 +50,21 @@ There is also another possible syntax to define stores. Similar to the Vue Compo
...
 ```
 
 In _Setup Stores_:
 
-- `ref()`s become `state` properties
-- `computed()`s become `getters`
-- `function()`s become `actions`
+- `ref()`s &nbsp;and `reactive()`s &nbsp;become `state` properties
+- `computed()`s &nbsp;become `getters`
+- `function()`s &nbsp;become `actions`
```
posva: let's revert all of these changes

pnpm-lock.yaml
```diff
@@ -92,7 +92,7 @@ importers:
       '@nuxt/module-builder': latest
       '@nuxt/test-utils': ^3.0.0-rc.8
       nuxt: ^3.0.0-rc.8
-      pinia: workspace:^2.0.18
+      pinia: '>=2.0.19'
```
posva: this needs to be reverted

packages/docs/getting-started.md
```diff
@@ -28,12 +28,12 @@ app.use(pinia)
...
 ```
 
-If you are using Vue 2, you also need to install a plugin and inject the created `pinia` at the root of the app:
+If you are using Vue 2, you also need to 'install' a plugin(`PiniaVuePlugin`) provided by pinia and inject the created `pinia` at the root of the app:
 
 ```js {1,3-4,12}
 import { createPinia, PiniaVuePlugin } from 'pinia'
 
-Vue.use(PiniaVuePlugin)
+Vue.use(PiniaVuePlugin) // installs the plugin
```
posva: let's revert this as well

packages/docs/core-concepts/actions.md
```diff
@@ -8,11 +8,12 @@
 Actions are the equivalent of [methods](https://v3.vuejs.org/guide/data-methods.html#methods) in components. They can be defined with the `actions` property in `defineStore()` and **they are perfect to define business logic**:
 
 ```js
-export const useStore = defineStore('main', {
+export const useCounterStore = defineStore('counter', {
   state: () => ({
     count: 0,
   }),
   actions: {
+    // use normal function syntax to define actions, arrow function syntax won't work
```
```suggestion
    // since we rely on `this`, we cannot use an arrow function
```

posva (changes requested):
Thanks!

packages/docs/getting-started.md
JeraldVin: including changes in line no. 31 @posva ?

packages/docs/core-concepts/index.md
JeraldVin: sure. Just for my information, is using reactive() to create state not encouraged @posva ?

pnpm-lock.yaml
JeraldVin: getting this error in build checks without this change(didn't manually change this. pnpm install did it)

 ERR_PNPM_OUTDATED_LOCKFILE  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with packages/nuxt/package.json

is it fine?

packages/docs/getting-started.md
```diff
@@ -28,7 +28,7 @@ app.use(pinia)
 app.mount('#app')
 ```
 
-If you are using Vue 2, you also need to install a plugin and inject the created `pinia` at the root of the app:
+If you are using Vue 2, you also need to 'install' a plugin(`PiniaVuePlugin`) provided by pinia and inject the created `pinia` at the root of the app:
```
posva: yeah, let's revert this too

packages/docs/core-concepts/index.md
```diff
@@ -105,15 +109,24 @@ export default defineComponent({
...
 
     return {
-      // will always be "eduardo"
+      // will always be "Eduardo"
       name,
-      // will always be 2
+      // will always be 0
       doubleCount,
-      // this one will be reactive
+      // will always be 0
```
```suggestion
      // will also always be 0
```

packages/docs/core-concepts/index.md
```diff
@@ -105,15 +109,24 @@ export default defineComponent({
...
+      doubleNumber: store.doubleCount,
+
+      // ✅ this one will be reactive
+      // will be 0 on initial render
+      // and will be 2 on re-render after setTimeout callback is executed
```
posva: Let's keep it simple
```suggestion
```

posva (changes requested):

packages/docs/core-concepts/index.md
posva: It's not encourage but supported

pnpm-lock.yaml
posva: No worries, I pushed a fix to the main branch

packages/docs/core-concepts/index.md
JeraldVin: can we have some note/tip on the same in the docs @posva ?

packages/docs/core-concepts/index.md
JeraldVin: changed to

`// will be 0 and then 2 after setTimeout callback is executed`

JeraldVin: did you mean to remove the lines or to simplify it @posva ?

packages/docs/core-concepts/index.md
```diff
@@ -105,15 +109,23 @@ export default defineComponent({
...
       name,
-      // will always be 2
+      // will always be 0
       doubleCount,
-      // this one will be reactive
+      // will also always be 0
+      doubleNumber: store.doubleCount,
+
+      // ✅ this one will be reactive
+      // will be 0 and then 2 after setTimeout callback is executed
```
```suggestion
```

posva (approved):
Thanks!

packages/docs/core-concepts/index.md
JeraldVin: @posva please consider mentioning in the docs that using reactive() is not encouraged, thanks!

PR #1581, opened 2022-08-23
docs(nuxt): pinia plugins usage (#1581) [skip ci]

JiatLn:
when use `export default myPlugin`, it got a nuxt3 warning following
```bash
[warn] [nuxt] You are using a plugin that has not been wrapped in `defineNuxtPlugin`. It is advised to wrap your plugins as in the future this may enable enhancements: myPlugin
```
it recommended to use
```ts
export default defineNuxtPlugin(({ $pinia }) => {
  $pinia.use(MyPiniaPlugin)
})
```

### DIFF
--- a/packages/docs/core-concepts/plugins.md
+++ b/packages/docs/core-concepts/plugins.md
@@ -368,7 +368,6 @@ When [using pinia alongside Nuxt](../ssr/nuxt.md), you will have to create a [Nu
 ```ts
 // plugins/myPiniaPlugin.js
 import { PiniaPluginContext } from 'pinia'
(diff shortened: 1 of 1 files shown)


posva (commented):
I think we might need to keep both versions in docs: one for Nuxt 2 and the other (definePlugin) for Nuxt bridge / Nuxt 3 🤔

posva (changes requested):
Can you add back the one that worked with Nuxt 2 to a new subsection named `### Nuxt.js 2`?

posva:
Thanks! I merged and did a few changes locally

Issue #859, opened 2021-11-30
Debugging information on Vue 2


posva:
Currently, the debugging information of events in the Timeline has almost nothing on Vue 2. While tha `events` array doesn't exist in Vue 2, there should be some kind of fallback to display at least some information like the previous values and current values for primitives.

### Discussed in https://github.com/posva/pinia/discussions/857

<div type='discussions-op-text'>

<sup>Originally posted by **Djaler** November 30, 2021</sup>
With the same code
![image](https://user-images.githubusercontent.com/7964583/143998537-9a8fd6cc-b02d-48be-99c6-045c26602bda.png)
mutation info in devtools in Vue 3 looks like
![image](https://user-images.githubusercontent.com/7964583/143998462-8eea6fb7-7929-4bfe-b5b9-791e71a8f442.png)
but in Vue 2 it's just
![image](https://user-images.githubusercontent.com/7964583/143998491-d56980ee-5522-4d2a-855f-cca508bc3d81.png)
 which is not so useful for debugging</div>

Djaler:
I think the only way we can do this is by saving previous store state deep clone

theolavaux:
Any updates on this topic? Do we know when the feature could be implemented? Keep up the good work 😄

stephane303:
I am missing completety  all the events for some of  my stores in Vue Devtools (6.1.3), which make the library very dangerous to use as it's almost impossible to track down the origin of the mutations.

parthibd:
I feel this is an important functionality that is currently missing from DevTools, any deadlines for its implementation?

parthibd:
I can try implementing this if someone can give me some hints on where to start looking into the code.

brandonburkett:
I actually do not see any timeline event information at all for pinia actions with vue 2.7.14 and invoked via `mapActions`.

PR #1585, opened 2022-08-23
feat(devtools): state diffing

KrosFire:
Hi inspired by redux devtools and also this issue: https://github.com/vuejs/pinia/issues/859

I've decided to add more debug information to pinia devtools about the status of the state before and after given action.
In acknowledgment that states can be pretty huge I've also decided to add `differences` object that only holds data that has been changed.

I think it works pretty well, but it will crash for any circular objects. On the other hand I don't think that those kinds of objects should stored in pinia state.

Let me know what you think :)

<details>
<summary>Here are some images of new devtools feedback:</summary>

Screenshot 2022-08-24 at 17 37 56
Screenshot 2022-08-24 at 17 38 12

Screenshot 2022-08-24 at 17 38 25
Screenshot 2022-08-24 at 17 38 39
</details>

### DIFF
--- a/packages/pinia/src/devtools/plugin.ts
+++ b/packages/pinia/src/devtools/plugin.ts
@@ -329,7 +329,7 @@ function addStoreToDevtools(app: DevtoolsApp, store: StoreGeneric) {
       store.$onAction(({ after, onError, name, args }) => {
         const groupId = runningActionId++
 
-        const initialState = copy(store.$state);
+        const initialState = copy(store.$state)
 
         api.addTimelineEvent({
           layerId: MUTATIONS_LAYER_ID,
--- a/packages/pinia/src/devtools/plugin.ts
+++ b/packages/pinia/src/devtools/plugin.ts
@@ -27,7 +27,7 @@ import {
   PINIA_ROOT_ID,
   PINIA_ROOT_LABEL,
 } from './formatting'
-import { isPinia, toastMessage } from './utils'
+import { isPinia, toastMessage, formatStateDifferences } from './utils'
 
 // timeline can be paused when directly changing the state
 let isTimelineActive = true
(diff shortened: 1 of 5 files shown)


packages/pinia/src/devtools/plugin.ts
```diff
@@ -328,6 +328,8 @@ function addStoreToDevtools(app: DevtoolsApp, store: StoreGeneric) {
       store.$onAction(({ after, onError, name, args }) => {
         const groupId = runningActionId++
 
+        const initialState = JSON.parse(JSON.stringify(toRaw(store.$state)));
```
posva: We should probably try to use https://developer.mozilla.org/en-US/docs/Web/API/structuredClone if it exists.
It should definitely gracefully fail for circular references and not crash

posva: why are you doing `toRaw()` here?

posva (changes requested):
This is pretty neat!
Make sure to run the lint:fix script in your files

Can you should some screenshots of the added features?

packages/pinia/src/devtools/plugin.ts
KrosFire: `toRaw` helps with cloning the object. I've tested `structuredClone` function but it seems it has problems with cloning functions.

I can create my own deepClone function or use the one that lodash provides, but I don't know if you want to have dependencies

posva: there shouldn't be functions in state 😓 
Yes, we want to avoid importing lodash

structuredClone have some limitations but it should also be quite fast. You should probably avoid doing `toRaw()` because we want to unwrap the properties.

I remember Vue devtools had a function for exactly this usecas in their codebase, it might be worth checking it out

KrosFire: From what I can tell, because state is wrapped in a proxy with methods, if I don't make this a raw object `structuredClone` won't work.

`JSON.parse(JSON.stringify)` will work without `toRaw` but the effect will be I think the same.

Could you explain to me why `toRaw` should be avoided in this situation?
I use this `raw` object just to copy and display it in devtools.

posva: Because of nested refs

KrosFire:
@posva I've added fast-copy util which is also used in vue devtools for vuex state copy. As far as I could see it all works great. I've attached some images of devtools.

posva:
Where is fast-clone used in devtools?
I think we might want to go a little bit more complicated with partial state patch objects so let's hold onto this feature for the moment

PR #1591, opened 2022-08-25
docs: fix a typo in the `pinia._StoreOnActionListenerContext.md` file

nazarepiedady:
Fellows, in this pull request, I send to merge a typing correction
that I did in the `pinia._StoreOnActionListenerContext.md` file.

It can seem an insignificant change, but necessary if we want to
keep a good and understandable documentation.

### DIFF
--- a/packages/docs/api/interfaces/pinia._StoreOnActionListenerContext.md
+++ b/packages/docs/api/interfaces/pinia._StoreOnActionListenerContext.md
@@ -72,7 +72,7 @@ ___
 ▸ (`callback`): `void`
 
 Sets up a hook if the action fails. Return `false` to catch the error and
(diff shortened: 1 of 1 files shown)


posva (changes requested):
Can you chagnge it in the source code instead?

posva (changes requested):
This file is generated, can you add it to the source code inside pinia instead?

nazarepiedady:
> This file is generated, can you add it to the source code inside pinia instead?

@posva I saw that the tsdoc is already updated, so the only thing missing is
to run the command to regenerate the API's markdown files.

I am closing this pull request because the task that needs to be done is simple
and you guys can do it anytime you want.

See you soon.

PR #1604, opened 2022-08-30
Added autoImportDir for stores/ dir if it exists and is specified 

richardeschloss:
Scope: Pinia's Nuxt module

This change adds the `autoImportDir` to the Nuxt module, to automatically import a user's "<srcDir>/stores" directory if it exists. This is a feature to help people who may wish to organize their stores into a separate directory, just like they did with Nuxt 2.x, but using the name "stores" to not conflict with the Nuxt 2.x directory "store".

I didn't have time tonight to provide an accompanying test, but if it this is a feature you think you'd like, please let me know and I can take a stab at writing the test. (fwiw, I tested manually in a test app that the feature works)

- [x] Auto importing of "stores" works
- [ ] Hot module reloading when stores changes...not quite there. Almost.


posva:
Is it really worth having an option for another option that only takes one line? I think it's better not to occult things from Nuxt users and that it's better if we teach them about the option and how to use it with pinia so they can use it with other libraries as well

richardeschloss:
> Is it really worth having an option for another option that only takes one line? I think it's better not to occult things from Nuxt users and that it's better if we teach them about the option and how to use it with pinia so they can use it with other libraries as well

After having given this some thought, I think it makes sense to keep one option `autoImports`, but have a check for individual entries that are directories. If a directory is encountered, use `addImportDir` instead. I thought it would be convenient to have the existence of "stores" folder be the thing that enables the feature (but I acknowledge there can be drawbacks with that).

richardeschloss:
Added tests, and refactored the code to use only the single `addAutoImports` option. If the option contains an existing directory, import it. Otherwise, import from composables.

posva:
@danielroe what do you think? I feel like obscuring the autoImport option from Nuxt like this for such a simple use case isn't a good idea.

danielroe:
Yes, if it's just adding user-configurable directories to autoimport array, we should instruct users to add them to their config using the Nuxt option, which seems simpler:

```js
  imports: {
    dirs: ['~/stores']
  }
})
```

There maybe is an option to import a directory like `~/stores` automatically (if it exists), though. And personally I would rather import `defineStore` and possibly some other imports automatically rather than requiring any user config.

packages/nuxt/__tests__/nuxt.spec.ts
```diff
@@ -27,3 +29,52 @@ describe('works with nuxt', async () => {
...
+        build: {
+          transpile: []
+        }
+      },
+      callHook() {},
+      hook(evt, fn) {
+        if (['autoImports:extend', 'autoImports:dirs'].includes(evt)) {
+          fn(_imports)
+        }
+      },
```
posva: Let's inline this variable in `nuxtCtx.set({ ... })`

posva (changes requested):
Okay so let's instead of allowing adding folders:

- add a `storesPath` option that defaults to a locally resolved `stores` path in the project
- add this to the documentation

This could also allow to maybe, in the future, allow domain-driven architectures where stores are nested in other folders: `*/stores`

posva:
@danielroe given that `defineStore()` is only used within stores, do you think it's really worth auto importing it?

I don't know if you (Nuxt) have benchmarked adding auto imports but I always wondered if there is a point where this affects performance

danielroe:
I want to do exactly that (benchmark auto-imports for larger projects). It's possible there's some opportunity to improve perf. Up to you, but I think it might be worth enabling it and leaving Nuxt to optimise any perf issues.

packages/nuxt/src/module.ts
```diff
@@ -1,11 +1,13 @@
...
  */
-import {
+ import { existsSync, statSync } from 'fs';
+ import {
   defineNuxtModule,
   addPlugin,
   isNuxt2,
-  addImports,
+  addAutoImport,
+  addAutoImportDir,
```
richardeschloss, line 10: `addImports` doesn't seem to be a supported type in @nuxt/kit yet.

packages/nuxt/src/module.ts
```diff
@@ -48,11 +54,11 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
   },
   defaults: {
     disableVuex: true,
-    autoImports: [],
+    autoImports: [
+      resolver.resolve('./stores')
```
richardeschloss, line 58: Let me know what you think of this idea as opposed to the "storesDir" option. The idea is, default `autoImports` to include the project's relative "stores" directory. Only if it exists will it be auto imported. 

This may provide some flexibility in case there are other directories the user may want to install.

packages/nuxt/src/module.ts
```diff
@@ -37,6 +42,7 @@ export interface ModuleOptions {
   autoImports?: Array<string | [string, string]>
 }
 
+const resolver = createResolver(import.meta.url)
```
richardeschloss, line 45: FYI: I moved the resolver statement to be in the outermost scope. Will this be ok?

danielroe: yes, it should be fine 👍

Rednas83:
I am a bit confused. Why can't all stores from the **stores** folder just be AutoImported just like with components/composables/content/pages/etc?

And when a custom folder is required just be added it to **Nuxt.config**. Something like:
```
  // https://pinia.vuejs.org/ssr/nuxt.html
  pinia: {
    autoImports: ['defineStore', 'skipHydrate', 'storesx/'],
  },
```

angelhdzmultimedia:
> Okay so let's instead of allowing adding folders:
> 
> * add a `storesPath` option that defaults to a locally resolved `stores` path in the project
> * add this to the documentation
> 
> This could also allow to maybe, in the future, allow domain-driven architectures where stores are nested in other folders: `*/stores`

Nice catch thinking about DDD! 

But this raised me a question, are Nuxt's auto imports directories taking this into consideration as well? You know, for consistence's sake...

And then I did a research and found this goodie https://dev.to/jacobandrewsky/using-modules-and-pinia-to-structure-nuxt-3-app-5963.

And that gave me an idea. If a project doesn't need DDD/DDA (Domain Driven Design/Domain Driven Architecture), 
then Pinia defaults the auto imports to root `~/stores` like Nuxt already does with everything else. Then when a project needs it, we use nuxt kit to create the modules and add the `*/stores` from each module to the autoImports array.

I'm new to all this so don't treat me harsh ha ha ha. **I decided to start learning Nuxt modules authoring today**. 🔥🎉🤯

_As a side note:_
I'm experimenting developing a tool like Nuxi to scaffold more templates using inquirer as "GUI" and it's working beautifully so far, to add stores, layouts, pages, server api, client api ($fetch or axios), entities (classes), types, interfaces, client plugins, server plugins, client middleware and server middleware. And when making the Pinia store template, I made  it with the `defineStore` import explicit. 

image

image


(if my experiment ends well, and I solve an issue I have with `unbuild` to generate the dist files, I might join Nuxt team to contribute to add these other templates to `Nuxi`).

### Final words
But if Pinia, besides adding the auto stores import,  also adds the hooks (defineStore, storeToRefs, skipHydrate, etc.)  to `autoImports` by default, I then can make the explicit imports optional for all the templates.

Thanks for opening this thread! Wishing you all health, peace, and success. 🔥🤝

angelhdzmultimedia:
Any updates?

I still find myself adding `stores/**` in imports map in the nuxt config.

posva:
Thanks for the PR. In the end, I decided to just add the auto imports automatically and remove the existing option.

angelhdzmultimedia:
Is this already pushed to main `v0.4.11`? Or will be done in the next version?

I removed `imports: { dirs: ['~/stores/**'] }` and it's complaining `useAuthStore is not defined`.

posva:
The default is different from yours. You can check the docs or the commit to see the diff

angelhdzmultimedia:
> The default is different from yours. You can check the docs or the commit to see the diff

Ooooh! I know what's happening! 🤯 Same as with Nuxt's composables.

If I want to have subdirectories, then I have to explicitly add the auto imports with the `dir/**` regex.
Your default is probably `stores` or `stores/*.ts`, so only .ts files in stores directory will be auto imported.

In my case, I have:
```ts
stores/
   auth/
      index.ts
      actions.ts
      state.ts
```

`stores/auth/index.ts`
```ts
import * as state from './state'
import * as actions from './actions'

export const useAuthStore = defineStore('auth', () => {
  return {
    ...actions,
    ...state,
  }
})
```

So yeah, I moved the files from stores/auth to composables/auth, and wrapped everything with reactive instead of using Pinia, to test, and it worked adding 
```
imports: {
    dirs: ['composables/**']
  },
   ```
   in nuxt config. So adding the same for Pinia stores dir will solve it, thanks! 🔥🤝

PR #1704, opened 2022-10-01
fix(store): init `_customProperties` for devtools

### LINKED ISSUE #1693: Production Build + Vue Devtool Plugin Error
### Reproduction

https://codesandbox.io/s/cedarexamplesandbox-forked-idhgg3?file=/vite.config.js

When attempting to use Vue Devtools in a production build by setting `__VUE_PROD_DEVTOOLS__: true`, there is an error when using Pinia

1. Go to https://codesandbox.io/s/cedarexamplesandbox-forked-idhgg3?file=/vite.config.js
2. Click on Terminal under preview pane
3. Click `+` to add another terminal instance
4. Execute a production build and start the server with `npm run build && npm run serve` (you should be prompted to open the new port in another browser pane)
image
5. Open the production build instance in a new window 
image
6. open the console and Vue devtools

### Expected behavior

Vue devtools should work as expected with no errors

### Actual behavior

Console error: 
```client-hook-6.js:1 TypeError: Cannot read properties of undefined (reading 'forEach')
    at index.b5e1fa3c.js:7:1055
    at addPlugin (VM240 backend.js:3068:7)
    at async Object.<anonymous> (VM240 backend.js:2229:5)
```
Looking through the stack trace, there is a [property that Pinia is adding in non-prod environments](https://github.com/vuejs/pinia/blob/v2/packages/pinia/src/store.ts#L454-L461), and is being referenced by the [devtools plugin code](https://github.com/vuejs/pinia/blob/v2/packages/pinia/src/devtools/plugin.ts#L386) 
image

### Additional information

_No response_

rudyxu1102:
Fix #1693

### DIFF
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -47,7 +47,7 @@ import {
   _StoreWithState,
 } from './types'
 import { setActivePinia, piniaSymbol, Pinia, activePinia } from './rootStore'
-import { IS_CLIENT } from './env'
+import { IS_CLIENT, USE_DEVTOOLS } from './env'
 import { patchObject } from './hmr'
 import { addSubscription, triggerSubscriptions, noop } from './subscriptions'
 
@@ -451,7 +451,7 @@ function createSetupStore<
 
   const store: Store<Id, S, G, A> = reactive(
     assign(
-      __DEV__ && IS_CLIENT
+      USE_DEVTOOLS
         ? // devtools custom properties
           {
             _customProperties: markRaw(new Set<string>()),
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -451,12 +451,16 @@ function createSetupStore<
 
   const store: Store<Id, S, G, A> = reactive(
     assign(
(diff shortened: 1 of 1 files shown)


posva (changes requested):
Thanks, there are other places where `_customProperties` are used that should also execute when adding devtools to prod

packages/pinia/src/store.ts
```diff
@@ -452,12 +452,16 @@ function createSetupStore<
   const store: Store<Id, S, G, A> = reactive(
     assign(
       __DEV__ && IS_CLIENT
-        ? // devtools custom properties
+        ?
```
posva: let's keep the whole thing in one ternary only. It's fine if `_hmrPayload` is there too

posva (changes requested):
There is also L671. Can you make sure that part runs with devtools as well?

packages/pinia/src/store.ts
rudyxu1102: Updated. Thanks.

posva (approved):
Thanks!

PR #1705, opened 2022-10-02
fix(doc): change the personal pronoun in doc

awxiaoxian2020:
The pinia has moved to vuejs, So I think using the third person is more reasonable.


awxiaoxian2020:
cc: @posva

posva:
thanks but this feels just weird as I wrote that part

PR #1709, opened 2022-10-05
fix(store): avoid prototype pollution

Debbl:
```js
for (const key in patchToApply) {
  if (!patchToApply.hasOwnProperty(key)) continue
  ...
}
```
```js
if (!patchToApply.hasOwnProperty(key)) continue
```
always return false,can't avoid prototype pollution

### DIFF
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -68,7 +68,7 @@ function mergeReactiveObjects<T extends StateTree>(
 
   // no need to go through symbols because they cannot be serialized anyway
   for (const key in patchToApply) {
-    if (!patchToApply.hasOwnProperty(key)) continue
+    if (!target.hasOwnProperty(key)) continue
     const subPatch = patchToApply[key]
     const targetValue = target[key]
     if (


posva (changes requested):
This breaks the existing behavior. Do you have a repro?

Debbl:
> This breaks the existing behavior. Do you have a repro?

> https://codesandbox.io/p/github/Debbl/pinia-template/draft/keen-river?file=%2Fsrc%2FApp.vue

App.vue file. The userStore does not have these attributes username and level, but it is working.

posva:
That's not a bug. It is however wrong usage. In any case, existing tests should still be passing.

Debbl:
> That's not a bug. It is however wrong usage. In any case, existing tests should still be passing.

```js
if (!patchToApply.hasOwnProperty(key)) continue
```
Thank you. I would like to ask if this line of code does not work under any circumstances.

posva:
It's for intherited properties: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

Debbl:
> It's for intherited properties: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

Thank you. It was my mistake.

PR #1715, opened 2022-10-11
Minor grammar updates to documentation

davidshq:
Some pretty minor changes to enhance grammar/clarity in `introduction.md`:
- "the initial principles are still the same" is now "the initial principles have remained the same"
- An instance of "pinia" was changed to "Pinia" to match the rest of the document's capitalization
- "Pina also support" is now "Pinia also supports"
- "through any RFC" is now "through any RFC process"
- "They were very often perceived" is now "They were often perceived"

### DIFF
--- a/packages/docs/core-concepts/index.md
+++ b/packages/docs/core-concepts/index.md
@@ -19,7 +19,7 @@ export const useStore = defineStore('main', {
 })
 ```
 
-This _name_, also referred as _id_, is necessary and is used by Pinia to connect the store to the devtools. Naming the returned function _use..._ is a convention across composables to make its usage idiomatic.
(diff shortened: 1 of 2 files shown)


davidshq:
Added a number of minor changes to core-concepts/index.md:
- "also referred as _id_" => "also referred to as _id_"
- Removed extra "and"
- Standardize on "Option stores" rather than "Options stores"
- "will get more complex [SSR]" => "will get more complex when using [SSR]"
- Standardize on "Options API" rather than "Option API"
- Capitalize Pinia
- "allow your bundler to code split and TypeScript inference" => "allowing your bundler to code split and providing TypeScript inference"
- "see" => "look at"
- "be just" => "just be"

posva (approved):
Thanks!

PR #1731, opened 2022-10-18
feat: warn developers for duplicated store keys

### LINKED ISSUE #1394: Warn if two stores are defined with the same name

rmartins90:
Another attempt to contribute to #1394 🙂

In tests, I'm getting various warnings because those tests are using duplicated names for stores. I'm not sure how to approach this in a smart way, appreciate any help or feedback.

Tried this approach to avoid messing with `._s` registry property. What are your thoughts?

Closes #1394

### DIFF
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -880,6 +880,14 @@ export function defineStore(
     id = idOrOptions.id
   }
 
+  // Warn developers about existing store ID in dev environments
+  if (__DEV__ && activePinia?._k.includes(id)) {
+    console.warn(`[🍍]: Stores should have unique identifiers. Found multiple stores with id "${id}". Rename one of them.`)
+  }
+
+  // Push ID to keys registry
+  activePinia?._k.push(id)
+
   function useStore(pinia?: Pinia | null, hot?: StoreGeneric): StoreGeneric {
     const currentInstance = getCurrentInstance()
     pinia =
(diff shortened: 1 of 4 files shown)


OXXXRA (approved):

packages/pinia/src/store.ts
```diff
@@ -880,6 +880,14 @@ export function defineStore(
     id = idOrOptions.id
   }
 
+  // Warn developers about existing store ID in dev environments
+  if (__DEV__ && activePinia?._k.includes(id)) {
+    console.warn(`[🍍]: Stores should have unique identifiers. Found multiple stores with id "${id}". Rename one of them.`)
+  }
+
+  // Push ID to keys registry
+  activePinia?._k.push(id)
```
andrewvasilchuk, line 889: I think you only need to push the key when `__DEV__` is true. There's no need to push it in production, as this will help reduce memory consumption

posva:
Sorry I never came back to this. I think this should be an eslint rule or similar. Otherwise, it will also warn during HMR

PR #1732, opened 2022-10-19
Delete anchor

isolcat:
## Description
Perhaps you forgot to remove the anchor when the Chinese document was officially launched, which made the Chinese document look very unattractive.
THE PROBLEMS happend at the doc: https://pinia.vuejs.org/zh/core-concepts/getters.html

I did not remove all the anchors, maybe you have a plugin that can completely remove all the anchors, I look forward to you being able to fix the Chinese documentation, thank you!

PR #1734, opened 2022-10-19
docs: remove anchor tag by markdown plugins.

ShenQingchuan:
Hi, I am the maintainer of Vite.js [docs-cn](https://github.com/vitejs/docs-cn). I found that the new Chinese translation of Pinia here these days. Nice job and shout out for @KimYangOfCat's great effort ❤️

I've noticed that there're still `{#xxx}` anchor tag remaining on `h1~h6` titles. **It is not a bug, it's a feature. we provide this mechanism in Vite's Chinese docs and Vuejs Chinese docs, and it's a general solution for anchor normalization.**

I'm here to help, and we just need two markdown-it plugins, as I submitted in this PR.

These two plugins is also used in the following repo:

**Vite Chinese docs**:
- https://github.com/vitejs/docs-cn/tree/main/.vitepress/render-perma-link/indes.js
- https://github.com/vitejs/docs-cn/tree/main/.vitepress/markdown-it-custom-anchor/index.js

**Vue.js Chinese docs**:
- https://github.com/vuejs-translations/docs-zh-cn/blob/main/.vitepress/headerMdPlugin.ts ( which has basically the same meaning )

Also, we can close #1732 after merging.

image

### DIFF
--- a/packages/docs/.vitepress/config.js
+++ b/packages/docs/.vitepress/config.js
@@ -1,5 +1,7 @@
 // @ts-check
 import locales from './locales'
+import renderPermaLink from './render-perma-link'
+import MarkDownItCustomAnchor from './markdown-it-custom-anchor'
 
 const META_URL = 'https://pinia.vuejs.org'
 const META_TITLE = 'Pinia 🍍'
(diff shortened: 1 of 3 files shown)


KimYangOfCat:
@ShenQingchuan  thanks for your PR, I am troubled by this problem when I was translating the chinese docs. I had suspected it‘s a bug, but I found that the vue Chinese documentation works very well, I knew I must miss something, but I have no time to keep looking into it.
thanks again!

KimYangOfCat (approved):

PR #1733, opened 2022-10-19
docs(zh): Polish the Simplified Chinese translations (Part 1)

awxiaoxian2020:
I will see also [translation guide for Vue](https://github.com/vuejs-translations/docs-zh-cn/wiki/翻译须知) to polish the translations.

PR #1750, opened 2022-10-25
docs(zh): word translation error

liangpengyv:
Original address in English: https://pinia.vuejs.org/core-concepts/#what-syntax-should-i-pick

Over here, "not sure" should not be translated as “不确实”，but as “不确定”.


KimYangOfCat:
thanks for your PR, But we haved fix this typo in #1733 . Please close this PR.

PR #1770, opened 2022-10-31
feat: resetAllStores helper for resetting all previously injects stores

Oleksii14:
For now, there is no possibility to reset all stores at once that were previously injected and used in the app. This PR introduces a new helper that allows resetting all the stores.

### DIFF
--- a/packages/pinia/__tests__/resetAllStores.spec.ts
+++ b/packages/pinia/__tests__/resetAllStores.spec.ts
@@ -1,0 +1,36 @@
+import { beforeEach, describe, it, expect, vi } from 'vitest'
+
+import { createPinia, defineStore, setActivePinia, Pinia } from '../src'
+import { resetAllStores } from '../src/resetAllStores'
+
+describe('createPinia', () => {
+  let pinia: Pinia
+
+  beforeEach(() => {
+    pinia = createPinia()
+    setActivePinia(pinia)
+  })
+
+  const useA = defineStore({
+    id: 'A',
+    state: () => ({}),
+  })
+
+  const useB = defineStore({
+    id: 'B',
+    state: () => ({}),
+  })
+
+  it('resets all stores', () => {
+    const stores = [useA(), useB()]
+
+    stores.forEach((store) => {
+      store.$reset = vi.fn()
+    })
+    resetAllStores()
+
+    stores.forEach((store) => {
+      expect(store.$reset).toHaveBeenCalled()
+    })
+  })
+})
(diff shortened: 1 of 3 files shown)


posva:
what is the use case? Why should this be in core?

IMO this kind of global helper should not be exposed in core because we don't want to encourage people to use the default `$reset()` behavior too much, it's more for debugging purposes. Resetting ALL stores looks more like a dangerous pattern as usually, only some stores are being reset at different moments in your application lifecycle (e.g. a store used for a form that spans over multiple pages)

Oleksii14:
> what is the use case? Why should this be in core?
> 
> IMO this kind of global helper should not be exposed in core because we don't want to encourage people to use the default `$reset()` behavior too much, it's more for debugging purposes. Resetting ALL stores looks more like a dangerous pattern as usually, only some stores are being reset at different moments in your application lifecycle (e.g. a store used for a form that spans over multiple pages)

As well as `$reset` that is available for the single store instance, there should be also a way to reset all stores at once.
Use case description: the user logged in to the application, visited some pages, and performed some actions and as a result, some pinia stores have now data saved in the state. The user logs out and the stores still contain some data. Another user (that may not have an access to some pieces of data) logs in with other credentials and can see the previously saved data in many stores. Of course, it can be fixed by manually resetting each store, but why if we can reset all of them using one function call after the logging out?

If you consider it to be a DANGEROUS pattern, can we just mark it as not recommended and describe the consequences?

posva:
> The user logs out and the stores still contain some data. Another user (that may not have an access to some pieces of data) logs in with other credentials and can see the previously saved data in many stores. Of course, it can be fixed by manually resetting each store, but why if we can reset all of them using one function call after logging out?

That is an app-specific design. Not all apps are built that way and even in that scenario resetting is completely overkill. It's also not explicit in its name. In your example, the `logout()` action should take care of all of that

Oleksii14:
> > The user logs out and the stores still contain some data. Another user (that may not have an access to some pieces of data) logs in with other credentials and can see the previously saved data in many stores. Of course, it can be fixed by manually resetting each store, but why if we can reset all of them using one function call after logging out?
> 
> That is an app-specific design. Not all apps are built that way and even in that scenario resetting is completely overkill. It's also not explicit in its name. In your example, the `logout()` action should take care of all of that

I got your point. Could you please add some technical details, why is that DANGEROUS or overkill?
Also, could you propose a better way to have all stores reset?

posva:
> Also, could you propose a better way to have all stores reset?

Keeping this exact same code in userland is totally fine!

Thanks for the PR! I'm closing this as I think it's better to keep it in userland for those few people who might need it.

dangerous because overkill because it resets every store instead of the ones that need it

Issue #991, opened 2022-01-25
subscriptions are lost during HMR


duoyue:
I have some code like this , it could output correctly when  app first loaded   or refresh ，but don‘t work after HMR. 

```javascript
const pinia = createPinia();
pinia.use(({ store }) => {
  store.$subscribe((mutations, state) => {
    console.log('trigger subscritions ', mutations, state);
  });
});
```

I have add some breakpoints in Pinia.js ，and found that `subscriptions ` is emty [] after HMR.
```javascript

// the  "subscriptions" is empty [] after HMR
function triggerSubscriptions(subscriptions, ...args) {
    subscriptions.forEach((callback) => {
        callback(...args);
    });
}

```

posva:
Maybe this can be fixed but it does require some extra logic because any subscription that happens inside a plugin should not be transferred. There could be other cases of subscriptions that should not be transferred to avoid duplication.

Or maybe there is even a better solution within `acceptHMRUpdate`

davidmyersdev:
@posva what do you mean by "any subscription that happens inside a plugin should not be transferred" here? I'm also running into this problem when trying to develop a plugin locally. The plugin only receives the `$subscribe` callback once before it is removed. Hot-reloading for plugins seems difficult since the initial plugin registration can have side effects on the current store state (e.g. hydrating the state from localStorage).

davidmyersdev:
Additionally, it looks like this problem goes away when using `detached: true` by preventing the [`removeSubscription`](https://github.com/vuejs/pinia/blob/ec7caa2d49a0ece2e1ed250aea2749dfdd6dd3b8/packages/pinia/src/subscriptions.ts#L22) callback from firing on unmount. Is that configuration expected for plugins that need to subscribe to mutations/actions?

Alanscut:
Passing `{ detached: true }` as the second parameter seems to fix this.

emrysctl:
thanks

zhu-hong:
When developing, pinia's hmr often doesn't work, which makes me very sick, either vite or vue cli

posva:
> Additionally, it looks like this problem goes away when using `detached: true` by preventing the [`removeSubscription`](https://github.com/vuejs/pinia/blob/ec7caa2d49a0ece2e1ed250aea2749dfdd6dd3b8/packages/pinia/src/subscriptions.ts#L22) callback from firing on unmount. Is that configuration expected for plugins that need to subscribe to mutations/actions?

[...]

PR #1781, opened 2022-11-07
docs(core-concepts): add info about using `detached: true` in store subscriptions inside plugins

F0rsaken:
Added note in docs about using `detached: true` with `$subscribe` and `$onAction` inside plugins in `Core conepts -> plugins` to help people avoid problems mentioned in this issue https://github.com/vuejs/pinia/issues/991 about losing subscriptions after a component is destroyed.

### DIFF
--- a/packages/docs/core-concepts/plugins.md
+++ b/packages/docs/core-concepts/plugins.md
@@ -189,6 +189,19 @@ pinia.use(({ store }) => {
 })
 ```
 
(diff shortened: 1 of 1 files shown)


posva:
Thanks for this. After some inspection I realized there is no problem, I even wrote a test for it in the subscriptions spec... I'm sorry I asked you to create a PR for this. The plugins in pinia are ran within the store's effect scope, which means they are disconnected from a regular `$subscribe()` in a component's setup...

I refactored the code at 57405de but it seems like it works without the change. If you found a case that doesn't work, a bug report with a reproduction would help.

PR #1808, opened 2022-11-16
fix(nuxt): nuxt v3 compatibility

pi0:
Update to RC.14 with latest api changes and dependency bumps (major) to prepare for v3.0.0

https://github.com/nuxt/framework/releases/tag/v3.0.0-rc.14

### DIFF
--- a/packages/nuxt/package.json
+++ b/packages/nuxt/package.json
@@ -45,14 +45,14 @@
     "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s --commit-path . -l @pinia/nuxt -r 1"
   },
   "dependencies": {
-    "@nuxt/kit": "^3.0.0-rc.13",
+    "@nuxt/kit": "^3.0.0-rc.14",
     "pinia": ">=2.0.23"
   },
   "devDependencies": {
-    "@nuxt/module-builder": "latest",
-    "@nuxt/schema": "^3.0.0-rc.13",
-    "@nuxt/test-utils": "^3.0.0-rc.13",
-    "nuxt": "^3.0.0-rc.13",
+    "@nuxt/module-builder": "^0.2.1",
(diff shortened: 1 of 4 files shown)


packages/nuxt/src/module.ts
```diff
@@ -5,7 +5,7 @@ import {
   defineNuxtModule,
   addPlugin,
   isNuxt2,
-  addAutoImport,
+  addImports,
   createResolver,
   resolveModule,
```
pi0: Note: `resolveModule` utility is deprecated from `@nuxt/kit` as is CJS-based resolution. But should be fine for pinia module.

posva: Should I use `const runtimeDir = fileURLToPath(new URL('./runtime', import.meta.url))` + `resolve()`?

pi0: We could replace it with [`mlly.resolve`](https://github.com/unjs/mlly#resolve) that implements ESM resolution and supports search path with `url: []` opt (similar to `import.meta.resolve`)

Simple `resolve`, does not support search path

pi0: BTW, it is nothing important IMO. We don't drop this util in Nuxt 3 only deprecating by type and current code works fine.

pi0:
Should be good to go @posva

packages/nuxt/src/module.ts
posva: ok!

brunokunace:
@pi0 have a schedule to create a new release with this version?

abdonrd:
I just tried the new [`@pinia/nuxt@0.4.4`](https://github.com/vuejs/pinia/blob/v2/packages/nuxt/CHANGELOG.md#044-2022-11-17), but it doesn't work:

https://github.com/nuxt/framework/discussions/9032#discussioncomment-4171427

PR #1818, opened 2022-11-21
fix(types): implemented a workaround to be TS 4.9.x compatible

jxn-30:
This commit fixes the issue mentioned in #1814.

I also had to update the `@microsoft/api-extractor` package because otherwise the build failed at step `pnpm run build:dts`.

I tested this workaround in a real life example and it worked as expected with both TypeScript versions, 4.9.3 and 4.8.4.

### DIFF
--- a/packages/pinia/package.json
+++ b/packages/pinia/package.json
@@ -69,7 +69,7 @@
   ],
   "license": "MIT",
   "devDependencies": {
-    "@microsoft/api-extractor": "7.28.6",
+    "@microsoft/api-extractor": "7.33.6",
     "@vue/test-utils": "^2.2.3"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


posva (approved):
Thanks!

elvisWEB:
After updating to 2.0.25, the build directly reported an error。。。

Charles1198:
> After updating to 2.0.25, the build directly reported an error。。。

error 是 `TS1005: '?' expected.` 嘛？

Charles1198:
> After updating to 2.0.25, the build directly reported an error。。。

Is the error like this `TS1005: '?' expected.` ？

jxn-30:
> After updating to 2.0.25, the build directly reported an error。。。

I think it would be better to put this into a new issue or discussion together with some usefull information (pinia Version, TS Version, OS, minimal reproduction code).

elvisWEB:
> 

yes~

elvisWEB:
> > After updating to 2.0.25, the build directly reported an error。。。
> 
> Is the error like this `TS1005: '?' expected.` ？

yes~

PasseRR:
waiting for fix compatibility.

PR #1822, opened 2022-11-23
fix(store): remove the state of disposed store

### LINKED ISSUE #1821: The new recreated store should not restore the last disposed store's state.
### Reproduction

https://stackblitz.com/github/DevilTea/repro-pinia-dispose-issue

1. Click the "Toggle" button on the left, the first time the store is installed. 
step 1

1. Click the "Increment" button on the right several times, and check the number of counts. 
step 2

1. Click the "Toggle" button on the left again, the store is disposed. 
step 3

1. Click the "Toggle" button on the left again, the store is installed, but its state is restored. 
step 4

### Expected behavior

The new recreated store should not restore the last disposed store's state.

### Actual behavior

The new recreated store restored the last disposed store's state.

### Additional information

I think the `$dispose` function should also delete the state inside `pinia.state.value`

https://github.com/vuejs/pinia/blob/v2/packages/pinia/src/store.ts#L342-L347

DevilTea:
closed #1821

### DIFF
--- a/packages/pinia/__tests__/store.spec.ts
+++ b/packages/pinia/__tests__/store.spec.ts
@@ -313,10 +313,11 @@ describe('Store', () => {
   })
 
   it('can be disposed', () => {
+    const initialN = 0
     const pinia = createPinia()
     const useStore = defineStore({
       id: 'main',
-      state: () => ({ n: 0 }),
+      state: () => ({ n: initialN }),
     })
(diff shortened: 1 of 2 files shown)


posva:
Thanks but see https://github.com/vuejs/pinia/issues/1821#issuecomment-1325334479

PR #1823, opened 2022-11-23
fix(nuxt): Avoid recursive import from Nuxt #imports

xel1045:
Since `runtime/composable.ts` is added to Nuxt `#imports`, we shouldn't use this import to define the composable. Without this fix, WebStorm is unable to analyze files importing Nuxt's `#imports` file and Vitest runs indefinitely.


posva (approved):
Thanks!

PR #1830, opened 2022-11-29
refactor: tidy up

Simon-He95:


packages/pinia/src/devtools/file-saver.ts
```diff
@@ -8,17 +8,17 @@
  */
 
 import { IS_CLIENT } from '../env'
-
+import { isPlainObject } from '../types'
```
posva, line 11: this file was copied so let's keep everything untouched

posva (changes requested):

PR #1842, opened 2022-12-06
fix(peer-deps): re-declaring peer-dependency of vue-demi

thebanjomatic:
When yarn installing this package, I receive the following error:
```
➤ YN0000: ┌ Resolution step
➤ YN0002: │ @pinia/testing@npm:0.0.14 [329d6] doesn't provide vue (pa561a), requested by vue-demi
➤ YN0000: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code
➤ YN0000: └ Completed
```
Which then yields the following message when running `yarn explain peer-requirements pa561a`
```
➤YN0000: @pinia/testing@npm:0.0.14 [329d6] doesn't provide vue, breaking the following requirements:
➤YN0000: vue-demi@npm:0.13.11 [e1ef3] → ^3.0.0-0 || ^2.6.0 ✘
```

It is customary to re-export the peer-dependencies of your dependencies so that consumers of @pinia/testing are able to identify what the peer dependency requirements are at a glance instead of needing to traverse the entire dependency graph to figure that out. The required version range here was taken from the latest release of vue-demi, but since the vue-demi dependency used here is *, it might be more appropriate to also just list the vue peer dependency version as *.

In the meantime, I was able to work around this issue by setting the following in my .yarnrc.yml:
```yaml
packageExtensions:
  "@pinia/testing@*":
    peerDependencies:      
      "vue": "*"

### DIFF
--- a/packages/testing/package.json
+++ b/packages/testing/package.json
@@ -50,7 +50,8 @@
     "siroc": "^0.16.0"
   },
   "peerDependencies": {
-    "pinia": ">=2.0.19"
+    "pinia": ">=2.0.19",
+    "vue": "^3.0.0-0 || ^2.6.0"
   },
   "publishConfig": {
     "access": "public"


posva:
> It is customary to re-export the peer-dependencies of your dependencies

That seems counterintuitive. Do you have a source? Is this yarn only? If a package dependency's peer dependencies are always needed, the package manager should always take care of it

thebanjomatic:
@posva this may be a yarn specific thing. I'm not sure if pnpm or npm have similar requirements as I don't personally use them.

Here is a reference from the yarn docs for the error message it prints on install: https://yarnpkg.com/advanced/error-codes#yn0002---missing_peer_dependency

>A package requests a peer dependency, but one or more of its parents in the dependency tree doesn't provide it.
>
>**Note that Yarn enforces peer dependencies at every level of the dependency tree**. That is, if ─D> is a dependency and ─P> is a peer dependency,
>
>```
># bad
>project
>├─D> packagePeer
>└─D> packageA
>     └─P> packageB
>          └─P> packagePeer
>
># good
>project
>├─D> packagePeer
>└─D> packageA
>     ├─P> packagePeer
>     └─D> packageB
>          └─P> packagePeer
>```

Which also links to the following blog post that further explains the problem: https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0

The blog post is a bit hand wavy in regard to the "isn't this something the package manager should handle?" argument, but I think the point is while presumably yarn could handle this, it doesn't currently, and it doesn't sound like they intend to.

thebanjomatic:
@posva any other thoughts or concerns with this PR?

thebanjomatic:
@posva its been a little while since I last commented so I thought I'd ping this PR again and see if you were ok with merging?

posva:
Thanks for the contribution, I found out this could break usage for existing users and creating really hard to debug errors. Usually one installs this package after installing vue and pinia so I don't think it adds much value to re declare the peer dependency until all package managers reach some agreement on how to make them work

Issue #1650, opened 2022-09-13
vue.js:5076 [Vue warn]: injection "Symbol(pinia)" not found.


youcanping:
### Reproduction

Reproduction

### Steps to reproduce the bug

- vue@2.7.x
- pinia@2.0.20

<img width="626" alt="截屏2022-09-13 16 25 02" src="https://user-images.githubusercontent.com/30832864/189850839-765fb640-9de9-45f0-bc3d-bfc5518f0edc.png">

console warn log: vue.js:5076 [Vue warn]: injection "Symbol(pinia)" not found.

### Expected behavior

no waring

### Actual behavior

warning 

### Additional information

after vue2.7

PR #1849, opened 2022-12-08
feat: avoid missing injection not found in edge Vue 2 edge case

XRenSiu:
issue #1650

### DIFF
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -882,21 +882,11 @@ export function defineStore(
 
   function useStore(pinia?: Pinia | null, hot?: StoreGeneric): StoreGeneric {
     const currentInstance = getCurrentInstance()
-    const getPiniaFromCurrentInstance = () => {
-      if (isVue2 && currentInstance && !currentInstance.$parent) {
-        const provides = currentInstance._provided
-        if (provides && piniaSymbol in provides) {
-          return provides[piniaSymbol]
-        }
-      } else {
-        return currentInstance && inject(piniaSymbol)
-      }
-    }
     pinia =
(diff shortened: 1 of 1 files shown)


posva:
This shouldn’t be necessary. Do you have a boiled down repro?

XRenSiu:
> This shouldn’t be necessary. Do you have a boiled down repro?

Reproduction: https://stackblitz.com/edit/vitejs-vite-t9sovg

posva:
Thanks!

The issue comes from the fact that `provide` doesn't work in the component passed to `new Vue()`, it does work in the one passed as `render: h => h(App)`. It counts as using the store globally, so passing the pinia instance will be necessary:

```ts
const app = new Vue({
  pinia,
  render: () => h(App),
  setup() {
    const store = useFooStore(pinia);
    const countStore = useCountStore(pinia);
    console.log(store.bar);
    console.log(countStore.count);
  },
});
```

However I think I have an idea to a simpler fix for this

packages/pinia/src/store.ts
```diff
@@ -882,11 +882,21 @@ export function defineStore(
...
+        }
+      } else {
+        return currentInstance && inject(piniaSymbol)
+      }
+    }
     pinia =
       // in test mode, ignore the argument provided as we can always retrieve a
       // pinia instance with getActivePinia()
       (__TEST__ && activePinia && activePinia._testing ? null : pinia) ||
-      (currentInstance && inject(piniaSymbol))
```
posva, line 889: Let's instead revert to the old version and give a default value to `inject(piniaSymbol, null)`. This should bypass the error and still work without false positives

posva (changes requested):

posva (approved):

PR #1850, opened 2022-12-08
feat(getters): add example with illegal(?) getter modification

Q-back:
Should it be illegal to modify a ref created from getter?

This PR is an example related to https://github.com/vuejs/pinia/issues/1851


posva:
Thanks for the clear example. This one should fail at the typing level (See https://github.com/vuejs/pinia/issues/1852) not allowing you to set a getter. If you want, you can give it a shot!

PR #1872, opened 2022-12-13
Fix code snippet error

Timibadass:
The current snippet implies there is another array inside the `autoImports` array and readers/users that copy and paste (like me) might have a hard time spotting  why their Store isn't working as expected. It also suggests that it is either `defineStore` or `'defineStore', 'definePiniaStore'` and not both.


posva:
Thanks but the array is normal and works in Nuxt 3

PR #1898, opened 2022-12-27
fix(types): type storeToRefs getters as ComputedRef

### LINKED ISSUE #1852: `storeToRefs()` should return `ComputedRef` for getters
Since getters cannot be written to, `storeToRefs()` should return a `ComputedRef` type while state properties return `Ref`

```ts
const useStore = defineStore('main', {
  state: () => ({ n: 0 }),
  getters: {
    double: state => state.n * 2,
  }
})

const { n, double } = storeToRefs(useStore())

double // ComputedRef<number>
n // Ref<number>
```

BenShelton:
Closes #1852

### DIFF
--- a/packages/pinia/src/storeToRefs.ts
+++ b/packages/pinia/src/storeToRefs.ts
@@ -1,14 +1,32 @@
 import {
+  ComputedRef,
   isReactive,
   isRef,
   isVue2,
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

PR #1906, opened 2022-12-31
docs(core-concepts): distinct methods of using store

lloydtao:
The docs reference three methods of using a store:
- Using the store instance
- Using computed properties
- Using `storeToRefs()`

These are blended together under "_Using the store_" and thus can be confusing for beginners.

I have split the methods into distinct subsections, as well as tweaking the code snippets in order to make the differences more obvious.


posva:
Thank you for this, I realise the changes are not out of date although the idea was good. Sorry I never managed to merge it 😓

PR #1914, opened 2023-01-03
Getters Chinese document translation defect

SailSpring:
Getters Chinese document translation defect. It should be consistent with English documents. It is ‘getters’ instead of ‘getter’.


awxiaoxian2020:
See also https://github.com/awxiaoxian2020/pinia/pull/1#discussion_r1002812260

PR #1978, opened 2023-02-02
docs(ja): add ja-JP translation

k-ta-yamada:
Hello.

I have translated the guide into Japanese for my own study as well.

The target is only the pages that seem to be frequently referred to in the guide,
but it may be a starting point for future work on the Japanese translation.

I would appreciate it if you could check and merge them.


posva (changes requested):
Hello! Thank you for this, could you move it outside of the repository like other translations? It might be worth moving to https://github.com/vuejs-translations/?q=ja&type=all&language=&sort= too

k-ta-yamada:
Thank you for your comment.
We will close this pull request, as the translated content may be out of date.

PR #1983, opened 2023-02-05
docs: playwright component testing instructions

sand4rt:


posva:
Thanks but I don’t think this should go in the official docs as it would also required to be maintained to be up to date 🙂
I would recommend you to write an article instead

PR #2002, opened 2023-02-15
docs(zh): update home page of the chinese docs and theme config

KimYangOfCat:
update home page of the chinese docs and theme config


Megasu:
useful~

PR #2024, opened 2023-02-24
docs: const name in the useMedia example

joaomelo:
The example code referenced a `video` variable that did exist in the example code (probably a copy-and-paste typo). The new code points to the `videoElement` const declared a few lines before.

Regarding "Please make sure to include a test! If this is closing an existing issue, reference that issue as well.", since it is a documentation update, I assume tests are not required.


posva (approved):
Thanks!

PR #2044, opened 2023-03-02
docs: import storeToRefs in example

mehicalen:
I noticed `storeToRefs` import is missing so this PR fixes that.


packages/docs/core-concepts/getters.md
```diff
@@ -98,6 +98,7 @@ and use in component:
 
 ```vue
 <script setup>
+import { storeToRefs } from 'pinia';
```
```suggestion
import { storeToRefs } from 'pinia'
```

posva (approved):
Thanks!

PR #2053, opened 2023-03-05
fix: allow `$reset` to be overridden by plugins

### LINKED ISSUE #2052: `$reset` cannot be overridden in plugins
### Reproduction

https://github.com/its-lee/pinia--reset-issue

After npm installing & npm run dev on the provided GitHub repository (https://github.com/its-lee/pinia--reset-issue)..

1. Click on the nonDollarAction button
2. Observe that in the console, that action invokation has led to the $onAction callback being run.
3. Click on the $reset button
4. Observe that in the console, the invokation of the $reset method *hasn't* led to the $onAction callback being run.

### Expected behavior

Both actions defined in defineStore and 'magic dollar' actions (notably $reset) should lead to the $onAction callback being invoked.

### Actual behavior

Only actions defined in defineStore appear to lead to the $onAction callback being invoked.

Any 'magic dollar' actions (notably $reset) do not lead to that occurring.

### Additional information

I encounted this issue when trying to write a plugin that followed this guide for adding new state (https://pinia.vuejs.org/core-concepts/plugins.html#adding-new-state). However, such state isn't reset by clients calling $reset.

The clear resolution was to add a $onAction callback in the plugin, that watched for when $reset would be called, and then led to the plugin resetting any state that it had added) - rather than doing something egregious like overwriting the generated $reset.

its-lee:
- Generates $reset earlier for stores created by the Options API, which allows plugins that introduce state to also include their state being reset when $reset is called (if desired). See https://github.com/vuejs/pinia/issues/2052 for where this was previously attempted. From reviewing the code, this would only affect plugins, giving them access to $reset which wasn't previously available, so wouldn't be a breaking API change.

```js
  const originalReset = store.$reset.bind(store) // Previously this wasn't available..
  return {
    $reset() { // In addition, previously the generation of $reset by pinia would have overwritten any new definition of a method called $reset created by a plugin.. 
      originalReset()
      // After effectively calling the original store.$reset() - reset state introduced by the plugin if needs be.
      store.secret = 'secret'
    }
  }
```

- Added tests for single & multi-plugin scenarios.
- Added documentation on the relevant $reset pattern that should be followed if necessary.

### DIFF
--- a/packages/docs/core-concepts/plugins.md
+++ b/packages/docs/core-concepts/plugins.md
@@ -160,6 +160,35 @@ pinia.use(({ store }) => {
 
 :::
 
+If you require that any state introduced by your plugin should be reset when `$reset` is called on your store, then you will need to overwrite `$reset` to include the resetting of your state:
+
+```js
+import { set, toRef } from '@vue/composition-api'
+pinia.use(({ store }) => {
+  if (!Object.prototype.hasOwnProperty(store.$state, 'hello')) {
+    const secretRef = ref('secret')
+    // If the data is meant to be used during SSR, you should
+    // set it on the `$state` property so it is serialized and
+    // picked up during hydration
+    set(store.$state, 'secret', secretRef)
+  }
+  // set it directly on the store too so you can access it
+  // both ways: `store.$state.secret` / `store.secret`
+  set(store, 'secret', toRef(store.$state, 'secret'))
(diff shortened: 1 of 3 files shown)


posva:
Thanks, I couldn't push to the branch so I created a copy at #2054

its-lee:
Thanks dude!

PR #2078, opened 2023-03-17
fix: support "types" condition in "exports" field

sheremet-va:
TypeScript with `node16` or `moduleResolution: "bundler"` cannot find types. See: https://arethetypeswrong.github.io


danielroe (approved):

posva (approved):
Thanks!

PR #2083, opened 2023-03-19
Remove reference to ViteSSR

stoivo:
ViteSSR are for Vite v2. Vite has build in support now https://vitejs.dev/guide/ssr.html

I think it shouldn't be here since it is for an old version of Vite. It was confusing to me

### DIFF
--- a/packages/docs/ssr/index.md
+++ b/packages/docs/ssr/index.md
@@ -78,33 +78,7 @@ app.use(pinia)
 devalue(pinia.state.value)
 ```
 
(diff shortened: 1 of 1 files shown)


posva:
Thanks, I think we should replace the example with something up to date rather than removing it

stoivo:
@posva, thanks for looking at it and your feedback. I removed it since it was pointing users to an outdated solution. I would not recommend anybody to go it that direction.

With the remaining documentation on this article and   vite ssr I manged to come up with a solution which worked for me. I'm not sure I would recommend this way but it's one way. I left some examples of how it can be done.

What do you think of this. You can alter the text if you want. Writing isn't my strong side.

posva (approved):
Thanks!

PR #2101, opened 2023-03-24
remove repetitive text 'API Documentation'

songbing97:
image


posva:
Thanks but this is auto generated and if removed it should be removed in the tool generation but it's not a big deal anyway

PR #2136, opened 2023-04-12
docs(zh): Update state.md

kongweigen:
There is something missing in the Chinese translation


KimYangOfCat:
This is because the Chinese documents are not synchronized in time.

LGTM

PR #2155, opened 2023-04-20
docs(zh): Update actions.md about Subscribing to actions describe

0x3E5:
Correction chinese expression.


KimYangOfCat:
LGTM @posva

PR #2163, opened 2023-04-21
docs: Update getters.md about parameter example

PineappleAttack:
Adding an example of a getter in a setup store with a parameter

### DIFF
--- a/packages/docs/core-concepts/getters.md
+++ b/packages/docs/core-concepts/getters.md
@@ -84,6 +84,8 @@ export const useCounterStore = defineStore('counter', {
 
 _Getters_ are just _computed_ properties behind the scenes, so it's not possible to pass any parameters to them. However, you can return a function from the _getter_ to accept any arguments:
 
+### Option Store
+
 ```js
 export const useStore = defineStore('main', {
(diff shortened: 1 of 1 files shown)


posva:
Thanks but already there: https://pinia.vuejs.org/core-concepts/getters.html#passing-arguments-to-getters

PineappleAttack:
> Thanks but already there: https://pinia.vuejs.org/core-concepts/getters.html#passing-arguments-to-getters

Sorry but I still see only an example for the option api not for setup variant.

hcmlopes:
I think this should be reopened. I just spent an hour googling and this example is not in the docs at all. 

There are some small nuances that should be in the docs for example your state being a `ref()`

PineappleAttack:
> I think this should be reopened. I just spent an hour googling and this example is not in the docs at all.
> 
> There are some small nuances that should be in the docs for example your state being a `ref()`

Thanks, I have not the rights to reopen it, but maybe @posva can do it?

posva:
In the future, I want to add a way to toggle between option and setup stores (like the vue docs)

PR #2167, opened 2023-04-24
feat(dx): throw an error if store id is missing

skirtles-code:
Calling `defineStore` without an id appears to succeed, but it fails when calling `useStore`:

```
Cannot read properties of undefined (reading 'startsWith')
    at devtoolsPlugin (pinia.mjs:925:19)
    at pinia.mjs:1631:48
    at EffectScope.run (reactivity.esm-bundler.js:37:24)
    at pinia.mjs:1631:38
    at Array.forEach (<anonymous>)
    at createSetupStore (pinia.mjs:1628:14)
    at createOptionsStore (pinia.mjs:1223:13)
    at useStore (pinia.mjs:1705:17)
```

The `undefined` id causes various other problems, but this just happens to be the first line that throws an error.

I've encountered this a few times on Vue Land, so I thought it might be better to have an explicit error in dev that explains what the problem is.

### DIFF
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -882,7 +882,7 @@ export function defineStore(
 
     if (__DEV__ && typeof id !== 'string') {
       throw new Error(
-        `[🍍]: defineStore must be passed an id string, either as its first argument or via the "id" option.`
+        `[🍍]: "defineStore()" must be passed a store id as its first argument.`
       )
     }
   }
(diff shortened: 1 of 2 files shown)


packages/pinia/src/store.ts
```diff
@@ -879,6 +879,12 @@ export function defineStore(
   } else {
     options = idOrOptions
     id = idOrOptions.id
+
+    if (__DEV__ && typeof id !== 'string') {
+      throw new Error(
+        `[🍍]: defineStore must be passed an id string, either as its first argument or via the "id" option.`
```
posva: I might deprecate the `id` option in the future since the `id` is always required anyway so I prefer to recommend only the argument way

```suggestion
        `[🍍]: "defineStore()" must be passed a store id as its first argument.`
```

posva (approved):
Thanks 🙌

PR #2190, opened 2023-05-09
docs: unify naming of store

Theiaz:
As [best practices describes](https://pinia.vuejs.org/core-concepts/#defining-a-store), stores should be named `useXXXStore`. This PR unifies the naming of `counter` store on the [testing](https://pinia.vuejs.org/cookbook/testing.html) section


posva (approved):
Thanks! For reference: you don't need to suffix `Store` every time. It sometimse makes things clearer but other times it doesn't add that much

PR #2195, opened 2023-05-10
fix(testing): clarify createSpy docs

leonheess:
The old wording made it sound like `vi.fn()` should be supplied when it should be `vi.fn` instead.

Additionally, there is super weird stuff happening if you supply `vi.fn()` so maybe that should be looked into at some point.


posva (approved):
Thanks!

PR #2199, opened 2023-05-13
chore: update `$state` JSDoc to reflect internal patching

Mini-ghost:
The purpose of this PR is to update the JSDoc of `$state` to reflect that when setting state, the state store will be patched via the `$patch()` method internally, rather than completely replacing the entire state store.

### DIFF
--- a/packages/pinia/src/types.ts
+++ b/packages/pinia/src/types.ts
@@ -321,7 +321,7 @@ export interface _StoreWithState<
   A /* extends ActionsTree */
 > extends StoreProperties<Id> {
   /**
-   * State of the Store. Setting it will patch the state via `$patch()` internally.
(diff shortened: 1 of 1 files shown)


packages/pinia/src/types.ts
```diff
@@ -321,7 +321,7 @@ export interface _StoreWithState<
   A /* extends ActionsTree */
 > extends StoreProperties<Id> {
   /**
-   * State of the Store. Setting it will replace the whole state.
+   * State of the Store. Setting it will patch the state via `$patch()` internally.
```
posva:
```suggestion
   * State of the Store. Setting it will internally call `$patch()` to update the state.
```

posva (approved):

PR #2256, opened 2023-06-12
chore: defining default values for Props using withDefaults

Alfred-Skyblue:
Define props default values using withDefaults.

### DIFF
--- a/packages/docs/.vitepress/theme/components/HomeSponsorsGroup.vue
+++ b/packages/docs/.vitepress/theme/components/HomeSponsorsGroup.vue
@@ -33,13 +33,15 @@ import sponsors from './sponsors.json'
 import { computed } from 'vue'
 import { useData } from 'vitepress'
 
-interface Props {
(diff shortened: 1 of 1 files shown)


packages/docs/.vitepress/theme/components/HomeSponsorsGroup.vue
```diff
@@ -31,22 +31,18 @@
...
-    type: [Number, String],
-    default: 140,
-  },
+interface Props {
+  name: 'Gold' | 'Platinum' | 'Silver' | 'Bronze'
+  size?: number | string
+}
+const props = withDefaults(defineProps<Props>(), {
+  size: 140,
 })
```
posva:
```suggestion
const props = withDefaults(
  defineProps<{
    name: 'Gold' | 'Platinum' | 'Silver' | 'Bronze'
    size?: number | string
  }>(),
  {
    size: 140,
  }
)
```

posva (approved):
Thanks!

PR #2260, opened 2023-06-14
docs(en): api call typo fix

terlankes:


posva:
Thanks but this is intentional for broader support: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

BTW, it's better if you open one PR with both Chinese and English changes, it's just easier to review 😄

PR #2261, opened 2023-06-14
docs(zh): api call typo fix 

terlankes:


posva:
See https://github.com/vuejs/pinia/pull/2260#issuecomment-1591308997

PR #2282, opened 2023-06-28
Explicit title for  store props destructuring

DespertaWeb:
I' m pretty sure lots of people are googling how to destructure the store. Let's make it easy with a nice and beautiful title


packages/docs/core-concepts/index.md
```diff
@@ -139,6 +139,7 @@ const doubleValue = computed(() => store.doubleCount)
 </script>
 ```
 
+## Destructuring store props
```
```suggestion
## Destructuring from a Store

```

PR #2292, opened 2023-06-29
fix(docs): remove invalid character from id.

ByMykel:
When I try to click on any element with the dolar sign I get the following error:

image
image

This is happening because the `id` attribute has an invalid character (`$`).

https://pinia.vuejs.org/api/interfaces/pinia._StoreWithState.html

### DIFF
--- a/packages/docs/api/interfaces/pinia.StoreDefinition.md
+++ b/packages/docs/api/interfaces/pinia.StoreDefinition.md
@@ -40,7 +40,7 @@ Returns a store, creates it if necessary.
 
 ## Properties %{#Properties}%
 
-### $id %{#Properties-$id}%
+### $id %{#Properties-id}%
 
 • **$id**: `Id`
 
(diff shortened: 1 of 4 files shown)


posva (changes requested):
Thanks but this are generated. They should be changed in the generation script in the docs package folde

ByMykel:
@posva ohh, should have taken more time to see it carefully. Can I still make the changes or is better to open an issue?

posva:
Make the changes but don’t open an issue about this

posva (changes requested):
I meant the changes to the script that generates the docs 😅

ByMykel:
Doesn't `packages/docs/typedoc-markdown.js` generate the id's that are failing?

posva:
Oh sorry, I didn't see on the phone. I'm starting to think this should actually be handled by vitepress 🤔 It should probably use `document.getElementById()` instead to avoid this specific problem. Vue Router does a similar thing

ByMykel:
@posva  I'm not that good to understand your solution, but I think `$` is not a valid character for an `id`. When I link you https://pinia.vuejs.org/api/interfaces/pinia._StoreWithState.html#Methods-$patch you can see that the page don't scroll down to `$patch` section.

So using `document.getElementById()` would fix the click, but not when opening the link.

Oscar-Raygoza:
oa

PR #2300, opened 2023-07-06
fix(devtools): Do not patch mocked actions

### LINKED ISSUE #2298: Pinia 2.1.4 devtools are incompatible with @pinia/testing
### Reproduction

https://github.com/bodograumann/pinia-testing-dev-tools-compat

1. Clone the repo
2. `npm install`
3. `npm run dev`
4. Open browser at http://localhost:5173

### Expected behavior

The testing pinia instance is used and you can see `Counter: 0` in the output.

### Actual behavior

Nothing is rendered and the browser console shows

> Uncaught TypeError: store.increment.mockImplementation is not a function

### Additional information

I am only using `@storybook/jest` to be able to have mocking while running in the browser. (For that to work I also had to set `window.global = window`.)
Maybe using a simpler spy creation method is also possible, but I would not know how right now.

bodograumann:
There were no devtools unittests afaics, so I added a new file.
Usually devtools are not loaded during unit tests, so I had to do the testing on the inner-most exposed level.
I am simulating a testing pinia with mocked actions and check that they are "sufficiently" preserved.

Fixes #2298

### DIFF
--- a/packages/pinia/__tests__/devtools.spec.ts
+++ b/packages/pinia/__tests__/devtools.spec.ts
@@ -1,0 +1,36 @@
+import { describe, it, expect } from 'vitest'
+import { mount } from '@vue/test-utils'
+import { createPinia, defineStore } from '../src'
+import { devtoolsPlugin } from '../src/devtools'
+
+describe('devtoolsPlugin', () => {
+  const useStore = defineStore('test', {
+    actions: {
+      myAction() {
+        return 42
+      },
+    },
+  })
+
+  it('preserves mocked actions during testing', () => {
+    const pinia = createPinia()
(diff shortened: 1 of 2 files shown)


doutatsu:
Any ETA on this getting merged?

bodograumann:
Anything I can do to get this merged, @posva ?
I really don't want to stay on an old version.
Does it make sense to rebase at this point?

posva:
This looks great, thanks for the PR! I haven’t been able to give this a look, sorry.
No need to rebase as it’s still mergeable

posva (approved):
Thanks and sorry for the delay!

PR #2327, opened 2023-07-26
pref:Remove set method return value in new Proxy

electroluxcode:
you know, the return statement inside the set method of a proxy doesn't affect the behavior of the proxy object. It is merely an ineffective statement and does not modify the property writing behavior. When using the set method, the main focus should be on processing the incoming property value, and there is no need to return any value.


posva (changes requested):
Does it really change anything in terms of Perf to not use `Reflect.set()`?

electroluxcode:
> Does it really change anything in terms of Perf to not use `Reflect.set()`?

I'm sorry, I confused the difference between the setter in proxy and the no-setter-return rule in ESLint.

PR #2337, opened 2023-08-01
VSCode to VS Code

burkeholland:
Small change to the name of VSCode > VS Code. Thank you for the great library!


posva (approved):
Thanks!

PR #2384, opened 2023-08-28
Update getting-started.md

zzuligy:
https://github.com/vuejs/vuex
vuex says that vuex and pinia can coexist, but here it says they can't. What's a more accurate way to say it?


posva (approved):
The note is specifically about v5 devtools so I think it's fine to remove it indeed. Thanks!

Issue #2388, opened 2023-08-30
Vue Demi missing export hasInjectionContext


acollins1991:
### Reproduction

https://github.com/acollins1991/pinia-vue-demi-bug/tree/main/friendly-sites-builder

### Steps to reproduce the bug

- Run `npm create vue@latest` to create a new vue project
- Run `npm i && npm run dev`
- Check browser console

### Expected behavior

No errors in the console

### Actual behavior

Prints the following error;

`Uncaught SyntaxError: The requested module '/node_modules/pinia/node_modules/vue-demi/lib/index.mjs?v=00078bdd' does not provide an export named 'computed'`

### Additional information

Fixed in https://github.com/vueuse/vue-demi/releases/tag/v0.14.6 release of Vue Demi

Tested update locally and doesn't break existing tests.

posva:
The fix was introduced earlier on in vue-demi. You need to update the version you have installed locally, you don't even need https://github.com/vuejs/pinia/pull/2390 to fix this

strokirk:
Anyone experiencing this issue while using Bun needs to add `vue-demi` to `trustedDependencies`, since otherwise [bun won't run](https://github.com/oven-sh/bun/issues/4959) the necessary `postinstall` scripts.

PR #2390, opened 2023-08-30
fix: missing vue-demi export hasInjectionContext, upgrade to v0.14.6

acollins1991:
https://github.com/vuejs/pinia/issues/2388


posva:
Thanks but see https://github.com/vuejs/pinia/issues/2388#issuecomment-1700661951

PR #2418, opened 2023-09-20
Ukrainian translation

matrunchyk:
Following https://github.com/vuejs/pinia/discussions/2223

Big big thanks to @maxdzin, @OlesyaKlochko, and @nazar1ua!

Just to note, everything is translated except interfaces - they will go in another PR.


packages/docs/.vitepress/config/index.ts
```diff
@@ -2,6 +2,7 @@ import { defineConfig } from 'vitepress'
 import { enConfig } from './en'
 import { sharedConfig } from './shared'
 import { zhConfig } from './zh'
+import { ukConfig } from './uk'
```
posva, line 5: Can you remove this? It's not needed since you are hosting the docs outside of this repo, which is easier to handle updates!

posva (changes requested):
Thanks a lot! Could you keep only the changes to the config/index.ts? Everything else is not needed

packages/docs/.vitepress/config/index.ts
matrunchyk: Makes send. Closing this PR and opened a new one with a singe change only https://github.com/vuejs/pinia/pull/2422

PR #2422, opened 2023-09-21
docs: Add Ukrainian language menu item

matrunchyk:
Following https://github.com/vuejs/pinia/discussions/2223

Big big thanks to @maxdzin, @OlesyaKlochko, and @nazar1ua!

Just to note, everything is translated except interfaces - they will go in another PR.

### DIFF
--- a/packages/docs/.vitepress/config/index.ts
+++ b/packages/docs/.vitepress/config/index.ts
@@ -24,5 +24,10 @@ export default defineConfig({
     	lang: 'pt-PT',
     	link: 'https://pinia-docs-pt.netlify.app/',
     },
+    uk: {
+      label: 'Українська',
+      lang: 'uk-UA',
+      link: 'https://pinia-ua.netlify.app/uk/',
+    },
   },
 })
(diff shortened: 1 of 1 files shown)


posva (changes requested):
One last thing: It would be better for the translation to not include other languages e.g. the Portuguese translation only has the Portuguese version https://pinia-docs-pt.netlify.app/

matrunchyk:
> One last thing: It would be better for the translation to not include other languages e.g. the Portuguese translation only has the Portuguese version [pinia-docs-pt.netlify.app](https://pinia-docs-pt.netlify.app/)

[Done](https://pinia-ua.netlify.app/)! Thanks for suggestions!

packages/docs/.vitepress/config/index.ts
```diff
@@ -24,5 +24,10 @@ export default defineConfig({
     	lang: 'pt-PT',
     	link: 'https://pinia-docs-pt.netlify.app/',
     },
+    uk: {
+      label: 'Українська',
+      lang: 'uk-UA',
+      link: 'https://pinia-ua.netlify.app/uk/',
```
posva:
```suggestion
      link: 'https://pinia-ua.netlify.app',
```

PR #2435, opened 2023-10-03
Do not add unofficial plugin

mikizdr:
# Description

If we want Pinia to be seriously taken, then not well tested and not supported by the wider community (consumers), then it should not be added. It creates confusion in the implementation and usage of Pinia.


posva:
Thanks for the concern but this is still helpful to some people.

PR #2452, opened 2023-10-16
fix: force active correctly scope in vue2

### LINKED ISSUE #1800: A getter not updated when option store used with Vue 2
### Reproduction

https://github.com/denego/piniajs-vue2-bug-report

Clone and start example project
```
git clone https://github.com/denego/piniajs-vue2-bug-report
cd piniajs-vue2-bug-report
npm i
npm run serve
```
Open http://localhost:8080
Click test2 link

### Expected behavior

`number2Getter` value matches `number2` value

### Actual behavior

`number2Getter` value stays the same

### Additional information

This project has two simple pinia stores, each has one number (number1 and number2), one getter (number1Getter and number2Getter) returning the number, and one action (updateTest1 and updateTest2) to increment the number.

There are two pages /test1 and /test2 each rendering TestCmp.vue. The component renders numbers and getters from both stores, and calls updateTest1 and updateTest2 on mount.

Initial render is good. However, if you click on test2 link to navigate, the mount hook of TestCmp.vue will be called, and number1, number1Getter, number2 will be incremented, but number2Getter which just returns number2, will not be updated.

Also interesting, if you place number2 and number2Getter before number1 and number1Getter in TestCmp.vue, then number1Getter will not be updated.

gcaaa31928:
This resolves the error scope issue caused on Vue 2 due to the callHook.

closed #1800

### DIFF
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -152,7 +152,10 @@ function createOptionsStore<
     if (!initialState && (!__DEV__ || !hot)) {
       /* istanbul ignore if */
       if (isVue2) {
+        const store = pinia._s.get(id)
         set(pinia.state.value, id, state ? state() : {})
+        // @ts-ignore: internal method
+        store?._s?.on();
       } else {
         pinia.state.value[id] = state ? state() : {}
       }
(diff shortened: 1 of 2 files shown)


posva:
See https://github.com/vuejs/pinia/issues/1800#issuecomment-1775774032

PR #2454, opened 2023-10-19
chore: add test for disposePinia

visualjerk:

### DIFF
--- a/packages/pinia/__tests__/lifespan.spec.ts
+++ b/packages/pinia/__tests__/lifespan.spec.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, vi } from 'vitest'
 import {
   createPinia,
   defineStore,
+  disposePinia,
   getActivePinia,
   setActivePinia,
 } from '../src'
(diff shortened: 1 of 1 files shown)


packages/pinia/__tests__/lifespan.spec.ts
```diff
@@ -168,4 +164,26 @@ describe('Store Lifespan', () => {
...
+    const pinia = createPinia()
+    setActivePinia(pinia)
+    const inStoreWatch = vi.fn()
+
+    const useStore = defineStore('a', () => {
+      watch(n, inStoreWatch, {
+        flush: 'sync',
+      })
+      return { n }
+    })
```
posva:
```suggestion
    const pinia = createPinia()
    setActivePinia(pinia)
    const inStoreWatch = vi.fn()

    const useStore = defineStore('a', () => {
      const n = ref(0)
      watch(n, inStoreWatch, {
        flush: 'sync',
      })
      return { n }
    })
```

we don't need to put the n outside in this case, it will only create a small leak in the tests

posva (approved):
Thanks!

PR #2459, opened 2023-10-22
fix: contrast problems with dark theme tip block in docs

sareg0:
While navigating the docs in dark mode I came across a 'Tip' block with a color combination that was difficult to read, yellow and white on yellow: 
Screenshot 2023-10-22 at 18 17 21

The above tip block can be found on the following page of the docs, while in dark mode: https://pinia.vuejs.org/core-concepts/state.html

This PR removes a custom theme value assignment that makes the tip's `background-colour` yellow.

## Local testing
Screenshot 2023-10-22 at 18 16 29

It's a bit better, in terms of contrast.


posva (approved):
Thanks!

PR #2473, opened 2023-11-02
docs: update the nuxt labs sponsor link

nazarepiedady:
I am just updating the **NuxtLabs** URL link that will send the user to their actual website.


posva:
Thanks! This is actually automatic from their profile. I need to refetch the whole file locally.

PR #2505, opened 2023-11-23
docs: fix command installation

hosseinmoghaddam:
I guess this is a page displayed on the Nuxt site, which causes problems for installation

### DIFF
--- a/packages/nuxt/README.md
+++ b/packages/nuxt/README.md
@@ -5,7 +5,7 @@
 ## Installation
 
 ```shell
-npm i @pinia/nuxt
+npm i pinia @pinia/nuxt
 ```
 
 ## Usage


posva:
Thanks. Why is this a problem? pinia is a dep of the nuxt package so it shouldn't be needed. I'm more than happy to remove it but it will be a breaking change and I remember I was told by the Nuxt team that this was the preferred approach for modules

hosseinmoghaddam:
I created a new Nuxt project just installed `@pinia/nuxt` and got this error.

image

Also, in the [Nuxt.js](https://pinia.vuejs.org/ssr/nuxt.html) part of Pinia's documentation, the following command is used to install the module:
`npm install pinia @pinia/nuxt`

posva (approved):

posva:
Oh this is the readme! I didn't see that. I guess pinia doesn't need to be a dependency then 😓

PR #2512, opened 2023-12-05
docs: add links to playground examples in introduction.md

skirtles-code:
This PR adds 4 links to `introduction.md`, linking to runnable examples.

Having examples like this throughout the docs would be really useful, but they also have the potential to be a maintenance hassle, e.g. ensuring the dependency versions are handled correctly. I figured it'd be better to start small and add more later.

The first two examples are almost exactly the same as the code shown in the docs. I think the only thing I changed was the imports.

The third and fourth examples needed more work, as they were incomplete.

I've copied some CSS from the core docs to show a Play icon next to the links. The original CSS is [here](https://github.com/vuejs/docs/blob/1350d08c364fc238994c8c77455766f6c41babb8/.vitepress/theme/styles/inline-demo.css#L1). The core docs are using a custom theme, so I had to make some adjustments to get it working in the Pinia docs.

I put the CSS in a separate file, `playground-links.css`. Not sure if that was the right decision, as it seems that most other CSS have been added to `vars.css`. If all custom CSS should go in `vars.css` then maybe it should be renamed?


posva:
This is a really good idea! I will take the time to review each example individually

posva (approved):
Thanks again!

PR #2535, opened 2023-12-21
Add vuemastery banner

i5dr0id:
image


packages/docs/.vitepress/theme/components/VueMasteryBanner.vue
```diff
@@ -0,0 +1,291 @@
...
+      </button>
+    </a>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { ref, onMounted } from 'vue'
+
+const isVisible = ref<Boolean>(true)
+const showFlash = ref<Boolean>(false)
```
posva:
```suggestion
const isVisible = ref(false)
const showFlash = ref(false)
```

BTW it's `boolean` in TS. `Boolean` is the constructor. Here the type is just inferred.

I'm also making it start as not visible. I should avoid a flash for users who already closed the banner.

PR #2541, opened 2024-01-02
chore: bump nuxt versions

danielroe:

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     ]
   },
   "resolutions": {
-    "@nuxt/kit": "^3.4.3"
+    "@nuxt/kit": "^3.9.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -68,6 +68,7 @@
     ]
   },
   "resolutions": {
-    "@nuxt/kit": "^3.4.3"
```
posva, line 71: I don't remember why this was needed on the first place but couldn't we just remove it now?

danielroe: you could. but sometimes renovate/dependabot end up installing multiple versions of these on upgrade prs and this helps make the upgrades smoother. your call though.

posva: I see, thanks for the info!

PR #2549, opened 2024-01-09
docs: setup translation infra

Jinjiang:
- contributing docs
- updated npm scripts
- added ~~2 script files~~ vitepress-translation-helper for comparing translation checkpoints and generating translation info (same as https://github.com/vuejs/router/pull/2109)
- added translation status UI
- upgraded vitepress to ~~rc35~~ rc36


awxiaoxian2020:
vitepress rc36 released 17 hours ago, should we upgrade to rc36?

Jinjiang:
> vitepress rc36 released 17 hours ago, should we upgrade to rc36?

I upgraded to rc35 just because it supports rtl direction from that version. For other versions I personally don't mind that. However, I think it's not necessary and it could be a separated PR.

Jinjiang:
@posva would you mind taking a look at this failing test, please? Since I think I didn't change anything related to this test but it failed after my commits. I also ran the same test locally and it succeeded. So I have no idea what is going on.

(and the rest of this PR is ready to review)

Thanks.

PR #2567, opened 2024-02-02
chore(docs): use local search

Sepush:
use local search

type define for local search
image
preview zh
image

preview en

image


posva:
Why? The Aloglia search works better

Sepush:
Aloglia is powerful,local is tiny.

PR #2578, opened 2024-02-13
fix: @pinia/nuxt breaks createTestingPinia

### LINKED ISSUE #2555: Pinia Nuxt module breaks the `createTestingPinia` helper in tests
### Reproduction

https://github.com/antoinezanardi/nuxt-pinia-testing

1. Clone the reproduction repository.
2. Run `pnpm install`
3. Run `vitest --config vitest.unit-config.ts`

### Expected behavior

The test should pass as the method from store is called when the component is mounted.

### Actual behavior

When using the `createTestingPinia` util method in a Nuxt project, the stores are not correctly mocked.

I created a simple test which expect a method from a store to be called when the component is mounted.

However, the test fails with the following logs :
``` bash
fetchAndSetRoles # <- This is the log from the store method, proving it's called and not correctly mocked.

stderr | tests/unit/app.nuxt.spec.ts > App Component > Store > should call method from store when mounted.

[Vue warn]: App already provides property with key "Symbol(pinia)". It will be overwritten with the new value. # <- This warning is not expected.

FAIL  tests/unit/app.nuxt.spec.ts > App Component > Store > should call method from store when mounted.

TypeError: [Function] is not a spy or a call to a spy! # <- Store is not mocked
```

### Additional information

I'm pretty sure that the nuxt module is the source of the problem.

If you comment the module in the `nuxt.config.ts` at the root of the project, the test passes.

```typescript
// nuxt.config.ts
  modules: [
    //"@pinia/nuxt", <- If you comment this line, the test passes …!
  ],
  devtools: { enabled: true }
})
```

OnlyWick:
fix #2555

### DIFF
--- a/packages/nuxt/src/runtime/plugin.vue3.ts
+++ b/packages/nuxt/src/runtime/plugin.vue3.ts
@@ -4,7 +4,11 @@ import { defineNuxtPlugin, Plugin } from '#app'
 
 const plugin: Plugin<{ pinia: Pinia }> = defineNuxtPlugin((nuxtApp) => {
   const pinia = createPinia()
-  nuxtApp.vueApp.use(pinia)
+
+  if (process.env.NODE_ENV !== 'test') {
+    nuxtApp.vueApp.use(pinia)
+  }
+
   setActivePinia(pinia)
 
   if (process.server) {


OnlyWick:
The reason is that there is a conflict between the pinia instance created by @pinia/nuxt and createTestingPinia. Ultimately, the pinia instance created by @pinia/nuxt takes precedence, causing the tests to fail.

packages/testing/src/testing.ts
```diff
@@ -106,7 +106,10 @@ export function createTestingPinia({
   fakeApp = false,
   createSpy: _createSpy,
 }: TestingOptions = {}): TestingPinia {
-  const pinia = createPinia()
+  const app: App = (globalThis as any).__unctx__.get('nuxt-app').tryUse().vueApp
+  const symbols = Object.getOwnPropertySymbols(app._context.provides)
+  const nuxtPinia: Pinia = app._context.provides[symbols[0]]
+  const pinia = nuxtPinia ?? createPinia()
```
posva: Thanks but not only we **always** want to create a new pinia instance here, this code is relying on internals and globals that are specific to nuxt. This will break in any other scenario

posva (changes requested):

packages/testing/src/testing.ts
OnlyWick: This is temporary fix code. You are right, this code lacks flexibility. However, there is another issue where `addPlugin` from `@pinia/nuxt` creates a new pinia instance, which conflicts with the creation of `createTestingPinia`. But the logic of `@nuxt/pinia` is reasonable, and I believe it should not be modified.

posva: It's okay to adapt `@nuxt/pinia` to work with the test environment. It's really crucial to always create a new pinia instance

OnlyWick: got it.

OnlyWick:
@posva pinia no export `piniaSymbol` symbol, so, we **cannot** reference pinia in the application by using `inject(piniaSymbol)`. And there is no example in the documentation that shows its usage in this way. It can also result in this problem occurring.

> app.provide(piniaSymbol, pinia)

```ts
install(app: App) {
  // this allows calling useStore() outside of a component setup after
  // installing pinia's plugin
  setActivePinia(pinia)
  if (!isVue2) {
    pinia._a = app
    app.provide(piniaSymbol, pinia)
    app.config.globalProperties.$pinia = pinia
    /* istanbul ignore else */
    if (__USE_DEVTOOLS__ && IS_CLIENT) {
      registerPiniaDevtools(app, pinia)
    }
    toBeInstalled.forEach((plugin) => _p.push(plugin))
    toBeInstalled = []
  }
}
```

OnlyWick:
Sad :-(. It's diffcult to me. I gave up.

I have subscribed to this issue, so please go ahead and fix it. I would like to learn how it is fixed.

posva:
No worries, thanks for the try!

PR #2588, opened 2024-02-22
refactor(nuxt): use object syntax for nuxt 3 plugin

huang-julien:
Hello this PR adds a name to the pinia nuxt plugin.

This will be necessary for plugins that relies on pinia, they'll be using the `dependsOn` property that require plugins to be named. In Nuxt 4, all plugins will be loaded in parallel by default


posva (approved):
Thanks!

alSergey:
@posva hi, can you please tell me when @pinia/nuxt will be released with this change? I'm really looking forward to them. I need the `dependsOn: ['pinia']` functionality, but in the current version I can't use it.

Neemsigo:
Hello, waiting for this feature too 🙂

PR #2604, opened 2024-03-09
fix(types): fix storeToRefs state return type (#2574)

### LINKED ISSUE #2574: `storeToRefs` loses the original typings of refs/computeds in a setup store, "generalizes" them instead
### Reproduction

https://play.pinia.vuejs.org/#eNqlV9ty2zYQ/RVUD5Xk2GSstJmpTDlOO0nGfUgysdI8RJmaIkEJNgVwAFC2R9W/9yxA0qTkOm7rF4GLvZy9Yr3pvS6KYF3y3rgXmUSLwjLDbVmwPJaLyaxnzax3OpNiVSht2YYZqzSfqk88M2zLMq1WrF8IKeL+SYurNPyCGBuWIExUKS3XgTX9mZzJMGTTpTBszbURSrJUccOkskzzLOeJZXbJWZzYMs6ZvStwqTK6M6Q7ZUI6BocmqLTFtm/YzfKOTS+YWaobw2LJuNZKMxjIheTsF/AmSho4SZLHbNJAHQzrqw2TYyaPgX3SdnfgRYbw0/Npjki9Jx3yOHAfuHrYsx86ZkePmB09bHY0ZDFcl9dS3Ug6nsuM6w9aLISMc/BNKUYRRQpx8iKnO0jJrBw1SKPQ5xvZxYflqyKPLccXY1HhfhgcG/tDNC+thTNnSS6Sa9SFPH72DJVxLhPNV1wCPvAjZtso9KyVhj3BKmqQ/USnDnsUesP39kf/aH/0gH0E74n2R4/Yj8ImGL3Dnq/po1VcBFdGSfTJhlhn1QXaY8wchWiuE4gy64XuHHCzOpprFCNK/wrMhzXrGZouTPnaKpWbo7gQXmxpbWHGYZikEvwpz8VaB5LbUBarcE/m7GXwc3Ac5mIewlAoZMpvu2YgcZTy1VO016xnz4Pjn4KXTuv6RaV0RVpJ6XYmtwiLNSitTCx2gpKoVSFyVGZhUfvd4MR5rm5+dzSrS95ATJY8uX6AfmVuPeqPlDO95i23bKwX3PrrNxfv+S3OzeVKpWUO7kcukXuVl4TRs/1ayhSwW3wO7bnLsZCLqXlza7mkhm6Aumg4fhfm3x5x/R7ui+BFK4r3QxEhbKZnyjMMq+4A9TO2NWJdHdt4Tr5krTnbodNQnkl+62T8LGhG86RtZ9B3M6N/yAZDNjn1yL2ABGdH6eA5ZhbdazwUWtLgcv4M4VCHccctTO9DN8tZGzEi10EpKCBZnBBXS1c0BYxY3p0ypJPL1JCSaFohdWbHFfa1Eik0AlKjNCtlQlnpOkLig3Wcl3zMpsPxnsHaklcL4+wVeMhnmrYkDTmast+xUiMfPGDjLwx1n4f0yZo6QpUHr/6PC+10o07rZ+UPUgyTXnhQPS7OHJtMJqxfw+xDqyODc+xP9EzWKvFSU8apjLLBrv7B0D1t/zn2ZObD/AorQ+BD8lGrgmt7N/BmD1nfBRGlXbVjlfFWodOf5w4q7/bCMPAO1R3PGJflimsCPGZZnJt6bt1o4dzoUv2wLPf4t6S21Useha9etJN/Q9JOI7kO2jDXtofsT/f7RdjlhcWLBcJneaPj4nWe44yhpbFEuau3aLcLWu0eGiztDtxs3Zez49aMrol3XHItEhNNHf0UwfKnJmc7mCIfAkGqKtbztIrLPdqouaeHt7quaO9QGtimusTXrvQM0aq3/lW9n56nKI7K0AlI0OgI+D2hzJIyECq1J7RnkiqQKqUnNNBIJTYyjkXOBSTlSR7DzUWu5thKXeWEBweO8YB9WXKJ4Yq3gpmCJ2Jvb8XK+pFi7SEajMIlF5qYIOPcgBbSr3laJthxrWIxPhYlrLJLlMVl4DfLJc9R4b4QEuy4c+6XYqsqLXDRrceqquDaCGEhOuHBJky8lcTBwWcTL/i4cefy8hK7jT/7Ls6Uqh8OWHtbfdV9Ucth+32DxHPtF7ddYfd/Q3e/ra9qRfcSYGzLBUT63upbC5x+7ePY/7aH7sMOso2ztLt0N6D+zdrd2K49QRD9KaSf+4baU7DTQE1jDXwVPqELv+IBB7n/zUv8+DSZqhe8FCWAxs72bwrIxQM=

Use `storeToRefs` on a setup store, containing special kinds of refs (see reproduction). The typings of the returned refs will be "generalized" to regular `Ref`s. The DX could be better here.

### Expected behavior

`storeToRefs` returns the original typings of the refs/computeds in the setup store.

### Actual behavior

`storeToRefs` returns generalized typings, even though the original refs are returned. So there's a discrepancy between what `storeToRefs` returns and how its return value is typed.

### Additional information

Read the comments in the reproduction to better understand the issue here. I included a workaround that kind of fixes the DX issue here, but it's probably not an ideal solution :wink:.

Note that I only tested refs, not computeds specifically.

nkeyy0:
Add new mapped type to keep original ref type in the `storeToRefs` return type (close #2574)

### DIFF
--- a/packages/pinia/src/storeToRefs.ts
+++ b/packages/pinia/src/storeToRefs.ts
@@ -11,7 +11,12 @@ import {
   toRefs,
 } from 'vue-demi'
 import { StoreGetters, StoreState } from './store'
-import type { PiniaCustomStateProperties, StoreGeneric } from './types'
+import type {
+  _UnwrapAll,
+  PiniaCustomStateProperties,
+  Store,
+  StoreGeneric,
+} from './types'
 
 type ToComputedRefs<T> = {
   [K in keyof T]: ToRef<T[K]> extends Ref<infer U>
@@ -20,12 +25,24 @@ type ToComputedRefs<T> = {
 }
 
 /**
+ * Extracts the refs of a state object from a store. If the state value is a Ref or type that extends ref, it will be kept as is.
+ * Otherwise, it will be converted into a Ref.
+ */
+type ToStateRefs<SS> =
+  SS extends Store<string, infer UnwrappedState>
+    ? UnwrappedState extends _UnwrapAll<Pick<infer State, infer Key>>
+      ? {
+          [K in Key]: State[K] extends Ref ? State[K] : Ref<State[K]>
+        }
+      : ToRefs<UnwrappedState>
+    : ToRefs<SS>
+
+/**
  * Extracts the return type for `storeToRefs`.
  * Will convert any `getters` into `ComputedRef`.
  */
-export type StoreToRefs<SS extends StoreGeneric> = ToRefs<
-  StoreState<SS> & PiniaCustomStateProperties<StoreState<SS>>
-> &
+export type StoreToRefs<SS extends StoreGeneric> = ToStateRefs<SS> &
+  ToRefs<PiniaCustomStateProperties<StoreState<SS>>> &
   ToComputedRefs<StoreGetters<SS>>
 
 /**
--- a/packages/pinia/src/storeToRefs.ts
+++ b/packages/pinia/src/storeToRefs.ts
@@ -12,8 +12,11 @@ import {
 } from 'vue-demi'
 import { StoreGetters, StoreState } from './store'
 import type {
+  _ActionsTree,
+  _GettersTree,
   _UnwrapAll,
   PiniaCustomStateProperties,
+  StateTree,
   Store,
   StoreGeneric,
 } from './types'
@@ -28,14 +31,19 @@ type ToComputedRefs<T> = {
  * Extracts the refs of a state object from a store. If the state value is a Ref or type that extends ref, it will be kept as is.
  * Otherwise, it will be converted into a Ref.
  */
-type ToStateRefs<SS> =
-  SS extends Store<string, infer UnwrappedState>
+declare type ToStateRefs<SS> =
+  SS extends Store<
+    string,
+    infer UnwrappedState,
+    _GettersTree<StateTree>,
+    _ActionsTree
+  >
     ? UnwrappedState extends _UnwrapAll<Pick<infer State, infer Key>>
       ? {
-          [K in Key]: State[K] extends Ref ? State[K] : Ref<State[K]>
+          [K in Key]: ToRef<State[K]>
         }
       : ToRefs<UnwrappedState>
-    : ToRefs<SS>
+    : ToRefs<StoreState<SS>>
 
 /**
  * Extracts the return type for `storeToRefs`.
--- a/packages/pinia/test-dts/customizations.test-d.ts
+++ b/packages/pinia/test-dts/customizations.test-d.ts
@@ -214,3 +214,23 @@ expectType<{
     })()
   )
 )
+
+expectType<{
+  n: Ref<number>
+  customN: Ref<number> & { plusOne: () => void }
+  double: ComputedRef<number>
+  myState: Ref<number>
+  stateOnly: Ref<number>
+}>(
+  storeToRefs(
+    defineStore('a', {
+      state: () => ({
+        n: 1,
+        customN: ref(1) as Ref<number> & { plusOne: () => void },
+      }),
+      getters: {
+        double: (state) => state.n * 2,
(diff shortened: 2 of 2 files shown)


posva (changes requested):
Thanks! Could you add a test for an options store using a special ref within the `state()` fn too?

nkeyy0:
> Thanks! Could you add a test for an options store using a special ref within the `state()` fn too?

Added

posva (approved):
Looks good! I need to test locally with the latest dependencies. Or if you have the time could you rebase against `v2` to get the latest deps?

nkeyy0:
Unfortunately, I won't have access to my laptop for the next 3-4 days and won't be able to rebase against v2 during this time. If it's urgent, feel free to handle it yourself. Otherwise, I can assist once I'm available. Let me know if there's anything else I can do

posva:
It's not urgent, don't worry 🙂

nkeyy0:
Initially, I forgot to handle the case with actions in the store, so I made some updates to the type and added tests for the store that contained the actions.

nkeyy0:
Also, I've rebased my feature branch onto `v2` and tested it locally. All good on my end. Let me know if you need anything else!

posva (approved):
Thanks!

packages/pinia/src/storeToRefs.ts
```diff
@@ -11,21 +11,46 @@ import {
...
   [K in keyof T]: ToRef<T[K]> extends Ref<infer U>
     ? ComputedRef<U>
     : ToRef<T[K]>
 }
 
+/**
+ * Extracts the refs of a state object from a store. If the state value is a Ref or type that extends ref, it will be kept as is.
+ * Otherwise, it will be converted into a Ref.
+ */
+declare type ToStateRefs<SS> =
```
nkeyy0, line 34: @posva Looks like `declare` was added accidentally here, but it still works. Should I add new PR to remove it? Or could you remove it yourself?

posva: I will remove it, thanks

posva:
Unfortunately, this wasn't fixing it; it set the returned type of storeToRefs to `any`, so I had to revert it. I should be able to give it more time in the following months or if any company really needs this, they can sponsor me and I will give it a higher priority

nkeyy0:
> Unfortunately, this wasn't fixing it; it set the returned type of storeToRefs to `any`, so I had to revert it. I should be able to give it more time in the following months or if any company really needs this, they can sponsor me and I will give it a higher priority

To be honest it's a bit unexpected. Could you provide cases when the return type is set to `any`?  Cause in type tests I didn't receive `any` return type. Maybe I just forgot to handle some specific case

posva:
It was unexpected to me too. See https://github.com/vuejs/pinia/commit/5c1c3e1d519c3b14b0f42e680e03441a00c5e886 and the failed checks in other commits.

nkeyy0:
Interesting that the initial commit was successful, but once you removed the `declare` it started to fail.
image

I've tested it locally and looks like [this commit](https://github.com/vuejs/pinia/commit/5c1c3e1d519c3b14b0f42e680e03441a00c5e886) broke the pipeline. It seems like adding `@internal` to JSDoc somehow changes the build process and `pinia.d.ts` is not generated correctly and I'm getting errors in type tests too. Could you try to remove `@internal` and give it one more try?)

nkeyy0:
Yeah, looks like adding `@internal` to `JSDoc` prevents emitting a declaration of `ToStateRefs` type and that's why we get `any` as a result. https://www.typescriptlang.org/tsconfig/#stripInternal, so removing `@internal` should fix the issue

nkeyy0:
@posva Just tagging you as hope this can still make it to release

PR #2616, opened 2024-03-16
docs: fix grammatical errors in getters.md

idorenyinudoh:

### DIFF
--- a/packages/docs/core-concepts/getters.md
+++ b/packages/docs/core-concepts/getters.md
@@ -133,7 +133,7 @@ const { getUserById } = storeToRefs(userList)
 </template>
 ```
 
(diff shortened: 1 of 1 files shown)


packages/docs/core-concepts/getters.md
```diff
@@ -133,7 +133,7 @@ const { getUserById } = storeToRefs(userList)
 </template>
 ```
 
-Note that when doing this, **getters are not cached anymore**, they are simply functions that you invoke. You can however cache some results inside of the getter itself, which is uncommon but should prove more performant:
+Note that when doing this, **getters are not cached anymore** — they are simply functions that you invoke. You can, however, cache some results inside of the getter itself, which is uncommon but should prove more performant:
```
```suggestion
Note that when doing this, **getters are not cached anymore**. They are simply functions you invoke. You can, however, cache some results inside of the getter itself, which is uncommon but should prove more performant:
```

PR #2617, opened 2024-03-17
docs: optimize content grammar in actions.md

idorenyinudoh:


packages/docs/core-concepts/actions.md
```diff
@@ -71,7 +71,7 @@ store.randomizeCounter()
 
 ## Accessing other stores actions
 
-To use another store, you can directly _use it_ inside of the _action_:
+To utilize another store, you can directly _get_ it inside of the _action_:
```
```suggestion
To consume another store, you can directly _use it_ inside of the _action_:
```

PR #2621, opened 2024-03-23
docs: fix and update module installation

selemondev:
This pull request is intended to fix and update the `pinia` and `@pinia/nuxt` module installation by fixing the installation commands. 

1. fix: The `pinia` package should be installed as a regular dependency while the `@pinia/nuxt` module should be installed as a dev dependency.
2. update: I have added `pnpm` among the package managers to be used when installing `pinia` and `@pinia/nuxt` module.

### DIFF
--- a/packages/docs/ssr/nuxt.md
+++ b/packages/docs/ssr/nuxt.md
@@ -5,9 +5,17 @@ Using Pinia with [Nuxt](https://nuxt.com/) is easier since Nuxt takes care of a 
 ## Installation
 
 ```bash
-yarn add pinia @pinia/nuxt
-# or with npm
-npm install pinia @pinia/nuxt
+# Using yarn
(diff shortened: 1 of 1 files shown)


posva:
Why the difference? What’s the bug?

selemondev:
> Why the difference? What’s the bug?

Hey @posva . The `nuxi` CLI, most of the Nuxt modules and the Nuxt devTools install nuxt modules ( including @pinia/nuxt ) as dev dependencies and not as regular dependencies. On the Pinia documentation site, the `@pinia/nuxt` module is installed as a regular dependency.

posva:
Thanks. I confirmed with Daniel that this is a thing of the past. Adding it as a dep is fine: https://github.com/nuxt/nuxt/issues/22050

PR #2622, opened 2024-03-25
fix: opt in to `import.meta.*` properties

danielroe:
This is a very early PR to make this module compatible with [changes we expect to release in Nuxt v5](https://github.com/nuxt/nuxt/issues/25323).

In [Nuxt v3.7.0](https://github.com/nuxt/nuxt/releases/tag/v3.7.0) we added support for `import.meta.*` (see [original PR](https://github.com/nuxt/nuxt/pull/22428)) and we've been gradually updating docs and moving across from the old `process.*` patterned variables.

As I'm sure you're aware, these variables are replaced at build-time and enable tree-shaking in bundled code.
This change affects _runtime_ code (that is, that is processed by the Nuxt bundler, like vite or webpack) rather than code running in Node. So it really doesn't matter what the string is, but it makes more sense in an ESM-world to use `import.meta` rather than `process`.

(It might be worth updating the module compatibility as well to indicate it needs to have Nuxt v3.7.0+, but I'll leave that with you if you think this is a good approach.)

### DIFF
--- a/packages/nuxt/src/runtime/plugin.vue2.ts
+++ b/packages/nuxt/src/runtime/plugin.vue2.ts
@@ -16,7 +16,7 @@ export default (context: any, provide: any) => {
     Object.defineProperty(store, '$nuxt', { value: context })
   })
 
-  if (process.server) {
+  if (import.meta.server) {
     context.beforeNuxtRender((ctx: any) => {
       ctx.nuxtState.pinia = pinia.state.value
     })
--- a/packages/docs/ssr/index.md
+++ b/packages/docs/ssr/index.md
@@ -92,7 +92,7 @@ const pinia = createPinia()
 const app = createApp(App)
 app.use(pinia)
 
-// `isClient` depends on the environment, e.g. on Nuxt it's `process.client`
+// `isClient` depends on the environment, e.g. on Nuxt it's `import.meta.client`
 if (isClient) {
   pinia.state.value = JSON.parse(window.__pinia)
 }
--- a/packages/nuxt/playground/app.vue
+++ b/packages/nuxt/playground/app.vue
@@ -8,7 +8,7 @@ useSomeStoreStore()
 
 // await useAsyncData('counter', () => counter.asyncIncrement().then(() => true))
 
-if (process.server) {
+if (import.meta.server) {
   counter.increment()
 }
 </script>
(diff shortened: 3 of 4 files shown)


posva (approved):
Looks good! I think it's fine not to add a peerdep version as it's  broken things with npm in the past and I think 3.7 was released a while ago

wJoenn:
@posva I think this PR broke pinia for Nuxt Bridge
```
 ERROR  in ./node_modules/@pinia/nuxt/dist/runtime/plugin.vue2.js                                                                                                                                                                                        friendly-errors 4:19:58 PM

Module parse failed: Unexpected token (15:12)                                                                                                                                                                                                            friendly-errors 4:19:58 PM
File was processed with these loaders:
 * ./node_modules/unplugin/dist/webpack/loaders/transform.js
 * ./node_modules/unplugin/dist/webpack/loaders/transform.js
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|     });
|   });
>   if (import.meta.server) {
|     context.beforeNuxtRender(function (ctx) {
|       ctx.nuxtState.pinia = pinia.state.value;
                                                                                                                                                                                                                                                         friendly-errors 4:19:58 PM
 @ ./.nuxt/index.js 40:0-70 300:23-50 305:17-44
 @ ./.nuxt/client.js
 @ multi ./node_modules/eventsource-polyfill/dist/browserify-eventsource.js (webpack)-hot-middleware/client.js?reload=true&timeout=30000&ansiColors=&overlayStyles=&path=%2F__webpack_hmr%2Fclient&name=client ./.nuxt/client.js
 ```
 
 Can conffirm reverting to `0.5.1` fixed the issue

File: packages/nuxt/src/runtime/plugin.vue2.ts [hunk 1]
```diff
@@ -16,7 +16,7 @@ export default (context: any, provide: any) => {
     Object.defineProperty(store, '$nuxt', { value: context })
   })
 
-  if (process.server) {
+  if (import.meta.server) {
```
posva, line 19: @danielroe Do you think we should just keep `process.server` for Vue 2? I suppose it won't be added to nuxt bridge

- https://github.com/vuejs/pinia/pull/2622#issuecomment-2304813676

danielroe: If it doesn't already, Nuxt bridge should support import.meta. If someone can open a bridge repro we can fix there.

wJoenn: I'll try to do that tomorrow if it hasn't been done already

PR #2625, opened 2024-03-27
chore: add getter double in playground demo-counter

FlyingCaiChong:
when I run playground, I found the code `<p>Counter :{{ counter.n }}. Double: {{ counter.double }}</p>`  in `DemoCounter.vue` , but in `demo-counter.ts` file, there is no `double`。so i think it's better to add `double` getter.


posva (approved):
Thanks!

PR #2634, opened 2024-04-03
Update index.md

HP8585:
in line 71 there were 2 mistakes , the first one is "Note you must" is absolutely wrong, it must be "Note that you must...", the second mistake is you called Pinia "pinia" in a non-code area which is wrong again...


posva (approved):
Thanks! FYI both can be used.
As a friendly recommendation for future contributions: you might want to communicate in more friendly/willing-to-help way rather than sounding criticizing ("is absolutely wrong", "you called [...] which is wrong again...") 😉

PR #2677, opened 2024-05-25
feat(docs): ensure lowercase for the anchor slug

nazarepiedady:
In this pull request, I am sending a change that will ensure that the anchor slug will be using lowercase letters.

`https://pinia.vuejs.org/introduction.html#Introduction` with this change will be `https://pinia.vuejs.org/introduction.html#introduction`.

### DIFF
--- a/packages/docs/.vitepress/config/shared.ts
+++ b/packages/docs/.vitepress/config/shared.ts
@@ -29,6 +29,8 @@ export const slugify = (str: string): string =>
     .replace(rSpecial, '-')
     // ensure it doesn't start with a number
     .replace(/^(\d)/, '_$1')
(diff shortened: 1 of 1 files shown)


posva:
Thanks but this is intended. hash can have upper case and it's needed for API.

BTW, for docs, use `doc:`. `feat` and `fix` are for source code only 🙂

nazarepiedady:
> Thanks but this is intended. hash can have upper case and it's needed for API.
> 
> BTW, for docs, use `doc:`. `feat` and `fix` are for source code only 🙂 

Oh, but the documentation of vue.js does not follow the same pattern. There, we have APIs with their names capitalised, but the hash is all in lowercase.

PR #2699, opened 2024-06-20
feat(nuxt): Support nuxt layers in a better way

marr:
When using the pinia nuxt module with nuxt layers, there were some things that don't appear to work correctly.

I am trying to solve for a nuxt layer workflow, which adds the pinia module and registers some stores in the base layer. I want those stores to be available to my containing app, as well as stores that the containing app defines.

The issues I found with the current state of the module include:

1. The pinia module tries to resolve _pinia/dist/pinia.mjs_ from the containing app. This is problematic, because the layer is the one bringing the pinia dependency in. It should be using the distributable that the layer depends on.
2. The stores directory in the layer gets lost when the containing app tries to register its own stores.
3. Extending the layer should bring in the stores from the layer, and allow adding stores from the containing app.

I believe this PR solves those three problems. We remove some use of a deprecated API as well.

To add tests to this, I think I'd need to create a new "playground-layer" directory and extend from that in a new nuxt spec pertaining to layer functionality. If you agree with these changes, we can discuss if you'd like that as part of this.

Note: This may break nuxt 2 compatibility. I can try to add back support if this is a decent enough approach to fixing layer support.

### DIFF
--- a/packages/docs/core-concepts/index.md
+++ b/packages/docs/core-concepts/index.md
@@ -127,7 +127,7 @@ You can define as many stores as you want and **you should define each store in 
 
 Once the store is instantiated, you can access any property defined in `state`, `getters`, and `actions` directly on the store. We will look at these in detail in the next pages but autocompletion will help you.
 
-Note that `store` is an object wrapped with `reactive`, meaning there is no need to write `.value` after getters but, like `props` in `setup`, **we cannot destructure it**:
+Note that `store` is an object wrapped with `reactive`, meaning there is no need to write `.value` after getters but, like any `reactive()` object in Vue, [**we lose reactivity when destructuring it**](https://vuejs.org/guide/essentials/reactivity-fundamentals.html#limitations-of-reactive):
 
 ```vue
 <script setup>


packages/nuxt/src/module.ts
```diff
@@ -97,7 +94,11 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
     }
 
     if (options.storesDirs) {
-      for (const storeDir of options.storesDirs) {
+      for (const storeDir of [
+        ...options.storesDirs,
+        /* @ts-expect-error storesDirs isn't on base NuxtOptions type */
+        ...(nuxt.options.pinia?.storesDirs || []),
```
posva: This doesn't seem necessary

posva (changes requested):
Thanks for the PR!
Adding the test would be a nice next step.
Regarding Nuxt 2, it should still work with nuxt bridge

packages/nuxt/src/module.ts
marr: I think this was added so that apps that extend layers using this module can still provide `storesDirs` that would be auto-imported. For example, if myapp had stores in a `mystores` directory, it could add those, while still importing stores the layer defines in its own context.

posva: If the layer adds options, I suppose they should appear in the `options` object. If not, it might be an intentional design for layers in Nuxt. Or are you referring to something else?

marr: Yes, thank you for correcting. I added tests and used explicit `storesDirs` in the layer config now to ensure that its stores get added.

marr:
@posva is this still something you might consider?

posva:
Yes! I will check it in my next pass

marr:
Thank you! If you are trying to reproduce the error you can follow these steps. Its also reproducible using a layer but you can also do it using a module like so:

```
npx nuxi create my-module -t module
// tweak package.json to add overrides: { vue: “latest” } — super annoying!!
npm install pinia @pinia/nuxt
// edit src/module.ts and simply `installModule(‘@pinia/nuxt’)` within the setup function
npm pack

// now to reproduce the error you can create a new nuxt project
npx nuxi init my-app
cd my-app
npm install ../my-module/my-module-1.0.0.tgz
// add `my-module` to the `modules` array in your app's nuxt.config.ts file.
npx nuxi prepare

 — see error below:

$ npx nuxi prepare

 ERROR  [unhandledRejection] Cannot find module 'pinia/dist/pinia.mjs'                                                                                                    1:52:52 PM
Require stack:
- /Path/to/my/index.js

  Require stack:
  - index.js
  at Module._resolveFilename (node:internal/modules/cjs/loader:1248:15)
  at Function.resolve (node:internal/modules/helpers:145:19)
  at Function._resolve [as resolve] (node_modules/jiti/dist/jiti.js:1:241814)
  at resolveModule (node_modules/@nuxt/kit/dist/index.mjs:2234:29)
  at setup (node_modules/@pinia/nuxt/dist/module.mjs:23:5)
  at normalizedModule (node_modules/@nuxt/kit/dist/index.mjs:2136:37)
  at async installModule (node_modules/@nuxt/kit/dist/index.mjs:2478:95)

✔ Types generated in .nuxt
```

Also note that this doesn't seem to happen using pnpm, so make sure to use npm for the reproduction.

posva:
The line that was problematic for layers has been removed. Could you confirm there is no need for something now?

marr:
> The line that was problematic for layers has been removed. Could you confirm there is no need for something now?

I am still experiencing an issue with `defineStore is not defined` when I have a layer that uses the module. The layer has `pinia` and `@pinia/nuxt` installed as dependencies. When the parent app loads the store that is exposed from the layer, it fails with that error.

I've tried installing the pinia packages in the parent app and registering the module there as well, but the error still occurs. It has been a problem for quite some time now.

posva:
This is still in my todo list. Good news is we don't need to support Nuxt bridge anymore. Nuxt 3 it is!

marr:
Note that this was solved via a couple ways:
1. I think there were recent updates to this module that fixed the layer lookup for pinia.
2. My layer was using a `srcDir`, which the pinia module doesn't account for when setting `storesDirs`. It is called out in the docs now (I'm not sure if it used to be?). The fix, anways, was to use a `createResolver` function to resolve the srcDir manually.

ex: in my layer/nuxt.config.ts...
```
import { createResolver } from '@nuxt/kit'

const { resolve } = createResolver(import.meta.url)

  srcDir: 'src',
  pinia: {
    storesDirs: [resolve('./src/stores/**')]
  },
...
```
That allowed the layer to find its own stores, while any apps that extend it can also set their own paths for the stores.

posva:
I see, thanks a lot!

mrleblanc101:
@posva @marr I see this has been merged in Pinia v2, will a new minor be released with this soon ?
Or will this come out with Pinia v3 ?
Or can @pinia/nuxt be updated with this feature independantly of pinia ?

PR #2717, opened 2024-07-15
docs: fix the Chinese api docs

### LINKED ISSUE #2716: docs: Chinese docs api documentation is 404
### Reproduction

https://pinia.vuejs.org/zh/api/pinia/

1. Open the pinia site and switch to Chinese
2. Click on the api document 
3. Click on the sidebar menu

image

### Expected behavior

Match to the correct route and markdown content

### Actual behavior

Now it jumps to a non-existent route and displays a 404

### Additional information

_No response_

zhangmo8:
image

close #2716

### DIFF
--- a/packages/docs/zh/index.md
+++ b/packages/docs/zh/index.md
@@ -6,7 +6,7 @@ titleTemplate: The intuitive store for Vue.js
 
 hero:
   name: Pinia
-  text: "得心应手的 \nVue.js 状态管理库"
+  text: "符合直觉的 \nVue.js 状态管理库"
   tagline: "类型安全、可扩展性以及模块化设计。\n甚至让你忘记正在使用的是一个状态库。"
   image:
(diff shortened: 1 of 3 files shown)


posva (changes requested):
Thanks! They files need to be updated instead of

packages/docs/zh/index.md
```diff
@@ -6,7 +6,7 @@ titleTemplate: The intuitive store for Vue.js
 
 hero:
   name: Pinia
-  text: "符合直觉的 \nVue.js 状态管理库"
+  text: "得心应手的 \nVue.js 状态管理库"
```
posva: Let’s revert this to the original

zhangmo8: Actually, I think it's better to be "得心应手"

posva (approved):

PR #2718, opened 2024-07-15
docs: fix case sensitive links

chriscpty:
The "Shared-Getters" and "Shared-Actions" links need to be capitalised to work - this is due to the markdown processor pinia uses giving them capitalised IDs (the "Shared Actions" heading is a ``<h2 id=Shared-Actions>`` rather than ``<h2 id=shared-actions>``

The proper fix for this would probably be to have the relevant markdown processor give headings lowercase IDs as that is more conventional afaik - this PR is meant as a stopgap to at least have the two links fixed in this commit work.

### DIFF
--- a/packages/docs/cookbook/composing-stores.md
+++ b/packages/docs/cookbook/composing-stores.md
@@ -42,7 +42,7 @@ const useY = defineStore('y', () => {
 
 ## Nested Stores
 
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks!

PR #2727, opened 2024-07-30
fix(nuxt): fix the import.meta ts error

zhangmo8:
Before:

image

After: 

image


posva:
Thanks but I don't have these errors and it's actually important to use `moduleResolution: Bundler` (default in nuxt)

Issue #2728, opened 2024-07-30
declare module 'vue' breaks types for auto imported components


J-Michalek:
### Reproduction

N/A

### Steps to reproduce the bug

1) Declare global component types under vue module:
```ts
declare module 'vue' {
  export interface GlobalComponents {
    ...
  }
}
```

2) Install pinia 2.2.0
3) Check for TS errors inside of Vue components that make us of the global components

### Expected behavior

It should not break the types for global components.

### Actual behavior

Breaks the types for global components

### Additional information

I am pretty sure this is the change that broke this in our project:
https://github.com/vuejs/pinia/commit/8a6ce86db83b6315c067c8a98c898b3c74efe62e#diff-094806c433011456914201ea662b339b53b9b1236f3226c1da3fd807106ad8a8R43

We use https://github.com/unplugin/unplugin-vue-components and the types it generates seemingly conflict with the ones Pinia declares.

posva:
In the ecosystem we were using `@vue/runtime-core` but the recommended way was to use `vue` and it took some time to adapt. At some point in time, upgrading TS versions and other dependencies **required** this change. In my testing, this change didn't break other libraries using `@vue/runtime-core` **when using up-to-date dependencies and dev dependencies**. You might need to update TS

mxp131011:
I've been in a similar situation

J-Michalek:
We are running on the latest TS version (5.5.4) but the issue is there after i update Pinia.

The only deps that are not up to date seem unrelated to this issue. I have listed them below.

dev deps:
- eslint
- eslint-plugin-chai-friendly

deps:
- vue-final-modal
- vuex

xmatthias:
I think the key here is the use in combination with unplugin-vue-components or unplugin-auto-import.

the generated component.d.ts file contains something around the following, then:

``` typescript
...
export {}

/* prettier-ignore */
declare module 'vue' {
  export interface GlobalComponents {
    MyComponent: typeof import('./components/MyComponent.vue')['default']
    MyOtherComponent: typeof import('./components/MyOtherComponent.vue')['default']
...
}
```

In my case - the update in the dependabot group broke all type-checks.
Reverting (and hard-pinning) pinia fixes this behavior again - with this, pinia is now the only dependency not uptodate.

**edit** while not a good, minimal reproduction (it's just my project breaking) - this PR showcases that the update of pinia breaks check-types (Ci is passing on main):
https://github.com/freqtrade/frequi/pull/2004 

[...]

PR #2730, opened 2024-08-06
fix(types): breaking type with auto imported components with Vue3

Bernankez:
### Related Issue

#2728 #2729 

### Reproduction

```sh
git clone https://github.com/Bernankez/pinia-types-repro.git
cd pinia-types-repro
pnpm install
```
Open `src/main.ts` and following the instructions.

### Environment

Node v20.15.0
VSCode v1.92.0
Vue.volar v2.0.28

### Additional information

I found that the import order of `router`, `App` and `pinia` can effect the `RouterView` type display in VSCode. If the `router` is imported first, then regardless of the order of `App` and `pinia`, the type of `RouterView` is always correct. If imported in the order of `App`, `pinia`, `router`, the type of `RouterView` will also be correct. However, if imported in the order of `pinia`, `App`, `router`, the type of `RouterView` will be `any`. I found that as mentioned in [this PR](https://github.com/unplugin/unplugin-auto-import/pull/469), adding an empty GlobalComponents export can fix this issue, regardless of the import order. I'm not sure if it's a pinia issue, so temporarily mark it as a draft.

### DIFF
--- a/packages/pinia/src/globalExtensions.ts
+++ b/packages/pinia/src/globalExtensions.ts
@@ -41,6 +41,7 @@ declare module 'vue/types/options' {
  */
 // @ts-ignore: works on Vue 3, fails in Vue 2
 declare module 'vue' {
+  export interface GlobalComponents {}
   export interface ComponentCustomProperties {
     /**
      * Access to the application's Pinia


posva (commented):
This is really curious, thanks a lot for opening the PR. I will give this a try, I wonder if it's because of the `export`... It feels so fragile 😓

posva (approved):
This seems to fix the issue with the order that doesn't work. Thanks again!

PR #2737, opened 2024-08-14
docs: More detailed explanation of `inject()` usage in `defineStore()`

Gianthard-cyh:
According to vuejs/core#11602, I believe users may confuse **globally provided properties** with **component-provided properties** when using Pinia. Therefore, I suggest emphasizing this in the documentation and including the implementation with Nuxt.

### DIFF
--- a/packages/docs/core-concepts/index.md
+++ b/packages/docs/core-concepts/index.md
@@ -99,6 +99,10 @@ export const useSearchFilters = defineStore('search-filters', () => {
 ```
 
 :::warning
+The injected properties must be **provided globally**. Component-provided properties cannot be accessed within the `defineStore()` function and will return `undefined` if injected.
+
+If you are using Nuxt, use `useNuxtApp().vueApp.provide('appProvided', 'value')` to provide a property globally.
+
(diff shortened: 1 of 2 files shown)


edison1105 (approved):

posva:
Thanks but this is already stated above. I find it redundant.
Regarding the Nuxt note, I don’t think this should be added here, people can look up the information in Nuxt docs

Anoesj:
> Thanks but this is already stated above. I find it redundant. Regarding the Nuxt note, I don’t think this should be added here, people can look up the information in Nuxt docs

I understand your position. Nuxt-specific info in the Pinia docs might get outdated and forgotten here. Instead of providing an example, we could link to this page in the Nuxt docs: https://nuxt.com/docs/api/composables/use-nuxt-app#provide-name-value.

posva:
The same argument applies to link to Nuxt docs: that page isn't related to nuxt and it shouldn't include _how to do app.provide in nuxt_

Anoesj:
Then maybe add this info here? https://pinia.vuejs.org/ssr/nuxt.html

posva:
It still doesn’t seem related…
I appreciate the proposal but I think it’s better not to merge this

PR #2740, opened 2024-08-15
chore: add `#__NO_SIDE_EFFECTS__` notation to `defineStore`

### LINKED ISSUE #2738: Tree-Shaking does not work when using multiple stores in Nuxt
> [!NOTE]  
> I’m not sure if this issue is related to Pinia, Nuxt, Vite or Rollup, but since the problem occurs specifically when using `defineStore` exports, and not with other composables or exports so I’m opening the issue here.
> 
> Please let me know if I should open it issue on the right repo.

### Reproduction

https://stackblitz.com/edit/github-dylldh?file=app.vue

1. Go stackblitz link above, run `nuxt generate`
2. Go `./output/_nuxt/Dz77aYJ1.js`
3. Search `foofoofoo`
4. If found, that means the store is not tree-shaken

pinia

### Bug Description

When using Pinia in a Nuxt project and creating more than one store, tree-shaking does not work as expected. Here's an example to illustrate the issue:

```ts
// store/foo.ts
export const useFooStore = defineStore('foo', () => ({ foo: ref('foofoofoo') }));

// store/bar.ts
export const useBarStore = defineStore('bar', () => ({ bar: ref('barbarbar') }));
```

```html
<!-- app.vue -->
<script setup lang="ts">
if (false) {
  // cannot tree-shake
  console.log(useFooStore());
  console.log(useBarStore());
}
</script>
```

After running `nuxt generate`, both the `foo` and `bar` stores are bundled into the output, even though they are inside a conditional block that should never run.

However, if I use only one store within the `if (false)` block, tree-shaking works as expected:

```html
<!-- app.vue -->
<script setup lang="ts">
if (false) {
  // can tree-shake
  console.log(useFooStore());
}
</script>
```

In this case, the unused store is successfully tree-shaken and not included in the bundle.

### Expected behavior

Tree-shaking should exclude unused stores from the final bundle, even when multiple stores are defined.

### Actual behavior

When more than one store is defined, tree-shaking fails to exclude unused stores, resulting in unnecessary code being bundled.

serkodev:
Fix #2738

In `defineStore`, certain behaviors of `createSetupStore` and `createOptionsStore` might prevent the bundler from performing static analysis to determine whether there are any side effects.

Therefore, in this PR, the `defineStore` function is marked as having no side effects, enabling the bundler to tree-shake stores that are not used after being defined.

## References
https://github.com/javascript-compiler-hints/compiler-notations-spec/blob/main/no-side-effects-notation-spec.md
https://github.com/rollup/rollup/pull/5024

### DIFF
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -870,6 +870,8 @@ export function defineStore<Id extends string, SS>(
   _ExtractGettersFromSetupStore<SS>,
   _ExtractActionsFromSetupStore<SS>
 >
+
+/*#__NO_SIDE_EFFECTS__*/
 export function defineStore(
   // TODO: add proper types from above
   idOrOptions: any,
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -870,7 +870,7 @@ export function defineStore<Id extends string, SS>(
   _ExtractGettersFromSetupStore<SS>,
   _ExtractActionsFromSetupStore<SS>
 >
-
+// improves tree shaking
 /*#__NO_SIDE_EFFECTS__*/
 export function defineStore(
   // TODO: add proper types from above


posva (approved):
Thanks a lot! So there is a way to improve it!

packages/pinia/src/store.ts
```diff
@@ -870,6 +870,8 @@ export function defineStore<Id extends string, SS>(
   _ExtractGettersFromSetupStore<SS>,
   _ExtractActionsFromSetupStore<SS>
 >
+
```
```suggestion
// improves tree shaking
```

PR #2757, opened 2024-08-31
Auto Register stores dir in all layers

OsamaHaikal:
Currently if user did not specify custom storesDir then the expected to auto import stores from stores dir but this is not working inside layers

1. Added automatic registration of the stores directory across all Nuxt layers.
2. Updated the documentation to clarify configuration handling with Nuxt layers.

### DIFF
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -67,8 +67,12 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
     ])
 
     if (!options.storesDirs) {
-      // resolve it against the src dir which is the root by default
-      options.storesDirs = [resolve(nuxt.options.srcDir, 'stores')]
+      // Add stores directory for each layer, including the main src dir
+      options.storesDirs = []
+      for (const layer of nuxt.options._layers) {
+        console.log(layer)
+        options.storesDirs.push(resolve(layer.config.srcDir, 'stores'))
+      }
     }
 
     if (options.storesDirs) {
--- a/packages/docs/ssr/nuxt.md
+++ b/packages/docs/ssr/nuxt.md
@@ -87,7 +87,7 @@ By default `@pinia/nuxt` exposes a few auto imports:
 - `storeToRefs()` when you need to extract individual refs from a store
 - `acceptHMRUpdate()` for [hot module replacement](../cookbook/hot-module-replacement.md)
 
-It also automatically imports **all stores** defined within your `stores` folder. It doesn't lookup for nested stores though. You can customize this behavior by setting the `storesDirs` option:
+It also automatically imports **all stores** defined within your `stores` folder in the main `srcDir` and all layers. It doesn't lookup nested stores though. You can customize this behavior by setting the `storesDirs` option:
 
 ```ts
 // nuxt.config.ts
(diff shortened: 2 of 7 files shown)


marr:
will this still register the stores dirs defined in layers if the main app defines it's own storesDirs? It seems like that null check would make it only apply the layers stores when the parent app sets it to null.

OsamaHaikal:
No because The `options` are resolved and merged from both the main app (layer) and all other layers. So If the user has defined storeDirs in either the main layer or any additional layer, there is no need to fallback to defaults, as this was the old behavior also

tommie:
This looks very good; thanks for writing the patch. I was just hitting this layer issue.

Why change the falsey check into a null check? If you want to keep it so, perhaps use `===`.

~~(I wonder why the second if-statement isn't an `else`. Might want to clean that up while at it.)~~ Never mind, it's building on the previous one.

tommie:
I made this wrapper module to work around this issue: https://gist.github.com/tommie/1e505f3f7a281698bdb9fdc7a303e6c0

OsamaHaikal:
Thank you 
> Why change the falsey check into a null check? If you want to keep it so, perhaps use `===`.
I thought that !options..storesDirs is not clear enough and since the need for the check for undefined and null i did ==

packages/nuxt/src/module.ts
```diff
@@ -91,9 +91,12 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
       { from: composables, name: 'storeToRefs' },
     ])
 
-    if (!options.storesDirs) {
-      // resolve it against the src dir which is the root by default
-      options.storesDirs = [resolver.resolve(nuxt.options.srcDir, 'stores')]
+    if (options.storesDirs == null) {
+      // Add stores directory for each layer, including the main src dir
+      options.storesDirs = []
+      for (const layer of nuxt.options._layers) {
```
posva, line 72: Let's use a public API for layers. I don't know what it is but surely there must be a way to resolve the layers from the options, maybe with a nuxt hook?

posva (changes requested):

packages/nuxt/src/module.ts
wzc520pyfm: `nuxt.options._layers` seems to be the currently recommended solution for nuxt at present. There are two reasons:
* This solution is introduced in the nuxt documentation. refer: [Multi-Layer Support for Nuxt Modules](https://nuxt.com/docs/guide/going-further/layers#multi-layer-support-for-nuxt-modules).
* In the nuxt source code, this api merging layer is also used. refer: https://github.com/nuxt/nuxt/blob/1c418d0ea3b046f2f7dedc3f39a87cc5be3e90b2/packages/nuxt/src/imports/module.ts#L62

packages/nuxt/playground/app.vue
```diff
@@ -6,6 +6,7 @@ const counter = useCounter()
 useTestStore()
 useSomeStoreStore()
 
+const layerStore = useLayerStore()
```
wzc520pyfm: I think we don't need to change the playground code. We can treat the current playground as a layer. Refer to #2828.
What do you think ?@posva

posva: Why not. I will need to take a proper look next time

lizyChy0329:
merge yet? It's work for me

```js
// base-layer
// "nuxt": "^3.15.0",
// "@pinia/nuxt": "0.9.0",
pinia: {
    // storesDirs: ['./stores/**'],
    storesDirs: [fileURLToPath(new URL('./stores/**', import.meta.url))],
  },
```

mrleblanc101:
Currently, I have a Layer with a dummy store, I had to add this to allow me to use any store define inside the layer otherwise it would only work with stores defined inside the playground.
```
pinia: {
    storesDirs: [join(currentDir, './app/stores')],
},
```
Then inside my app extending from this layers, my stores wouldn't work anymore, I suppose because the `storesDirs` defined in the layer overwrite the default value inside the app, so I added this to my app:
```
pinia: {
    storesDirs: ['./app/stores'],
},
```
Now, if I create a store with the same name in my app as in the layer, the store defined in the app will overwrite the dummy store inside the layer, this now work exactly like components overwritting between app and layer 👌

I hope this gets merged

OsamaHaikal:
Updated to point to the new default branch

LasseSLambertsen:
So any news on when this will be merged? :)

alectrocute:
How is it October already?

PR #2759, opened 2024-09-02
Docs: Migrate mockedStore typings to vitest v2

lud-hu:
Update docs to migrate the typings of `mockedStore` to the changes of >vitest@2.0.0:

see:
https://vitest.dev/guide/migration#simplified-generic-types-of-mock-functions-e-g-vi-fn-t-mock-t
https://github.com/vitest-dev/vitest/releases/tag/v2.0.0

### DIFF
--- a/packages/docs/cookbook/testing.md
+++ b/packages/docs/cookbook/testing.md
@@ -191,7 +191,7 @@ function mockedStore<TStoreDef extends () => unknown>(
           ...args: infer Args
         ) => infer ReturnT
           ? // 👇 depends on your testing framework
-            Mock<Args, ReturnT>
+            Mock<(args: Args) => ReturnT>
           : Actions[K]
       }
     > & {


posva (approved):
Thanks!

packages/docs/cookbook/testing.md
```diff
@@ -191,7 +191,7 @@ function mockedStore<TStoreDef extends () => unknown>(
           ...args: infer Args
         ) => infer ReturnT
           ? // 👇 depends on your testing framework
-            Mock<Args, ReturnT>
+            Mock<(args: Args) => ReturnT>
```
posva, line 194: Although now it should be able to just take `Actions[K]`. I need to test it

posva (changes requested):

packages/docs/cookbook/testing.md
netroy, line 194: Shouldn't this be

```suggestion
            Mock<(...args: Args) => ReturnT>
```

PR #2782, opened 2024-09-30
docs: update contributing guide

jh-leong:
Update the repository link and recommend using the official Pinia playground for bug reports instead of jsfiddle.net :)


posva (approved):
Thanks!

PR #2793, opened 2024-10-04
feat(hmr): extend `acceptHMRUpdate` to allow for cleaning up side effects of existing store

Anoesj:
### Changes
This PR adds to possibility of passing a cleanup function as an optional third argument to `acceptHMRUpdate`. This cleanup function can be used to clean up side effects of the existing store before initializing the new store. This is useful if you have event listeners or other side effects that need to be cleaned up in the old store.

This solves a non-trivial case. Usually, having side effects inside a Pinia store is probably not such a great idea. However, it happens, and if it does, we need to provide a way to make sure those side effects can be properly cleaned up during HMR.

Let's say you have an app that relies on a lot of event listeners, and you want to manage them centrally and close to some global state. If there isn't an obvious Vue component to add and remove all these event listeners, you might choose to manage these event listeners in a Pinia store. While working on this Pinia store, you might need to clean up some of those event listeners, before HMR replaces the existing store with the new one. We can use `onScopeDispose` to run some code when the store is disposed, but nothing is disposed when HMR swaps the stores, so we can't use that mechanism here. For better control over HMR, we need a way to talk to the existing store, just before it gets replaced. This PR provides that mechanism.

### To do
- **Tests:** I couldn't find a way to properly test this. If someone can assist with that, that would be nice! I've manually tested using the playground (see the `/scroll-store` route).
- **Translations:** I updated the English docs about HMR, but the Chinese version needs to be updated as well.
- **Explore alternatives:** I can imagine something like an `onBeforeHMRReplace` hook being a viable alternative. A user would need to call `onBeforeHMRReplace` in a setup store, just like with `onScopeDispose`. It would probably require us to detect and register any called `onBeforeHMRReplace` hooks during the setup phase (store initialization) and manually trigger their callbacks in `acceptHMRUpdate`.

### Example (taken from the modified docs)
```js
// scroll.js
import { defineStore, acceptHMRUpdate } from 'pinia'
import { ref } from 'vue'

export const useScroll = defineStore('scroll', () => {
  const scrollTop = ref(window.scrollY)

  function onScroll () {
    scrollTop.value = window.scrollY
  }

  function trackScroll () {
    window.addEventListener('scroll', onScroll, { passive: true })
  }

  trackScroll()

  function $cleanUp () {
    window.removeEventListener('scroll', onScroll)
  }

  return {
    scrollTop,
    trackScroll,
    $cleanUp,
  }
})

if (import.meta.hot) {
  import.meta.hot.accept(acceptHMRUpdate(useScroll, import.meta.hot, (existingStore) => {
    existingStore.$cleanUp()
  }))
}
```

### DIFF
--- a/packages/docs/cookbook/hot-module-replacement.md
+++ b/packages/docs/cookbook/hot-module-replacement.md
@@ -2,8 +2,8 @@
 
 Pinia supports Hot Module replacement so you can edit your stores and interact with them directly in your app without reloading the page, allowing you to keep the existing state, add, or even remove state, actions, and getters.
 
-At the moment, only [Vite](https://vitejs.dev/guide/api-hmr.html#hmr-api) is officially supported but any bundler implementing the `import.meta.hot` spec should work (e.g. [webpack](https://webpack.js.org/api/module-variables/#importmetawebpackhot) seems to use `import.meta.webpackHot` instead of `import.meta.hot`). 
-You need to add this snippet of code next to any store declaration. Let's say you have three stores: `auth.js`, `cart.js`, and `chat.js`, you will have to add (and adapt) this after the creation of the _store definition_:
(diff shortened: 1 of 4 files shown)


posva (changes requested):
I think we should simply dispose of the old store to trigger `onScopeDispose()` just like components trigger `onUnmounted()`. There could be limitations but I'm more keen on exploring that side.

PR #2794, opened 2024-10-06
fix(types): handle union types in generic parameter

### LINKED ISSUE #2785: Common state between stores is lost if type is different
### Reproduction

https://play.pinia.vuejs.org/#eNqtVUtv20YQ/isDXkgDElnXRQqokmMnTdH2kBa2L0GYA0WOxLXIXWJ3SSkQ9N8zu8un/EgOsSGC3Pnmm8fO4+jdVlXY1OgtvKVKJas0KNR1BUXCt6vY0yr2rmPOykpIDUeoFd5rIRFOsJGiBD+MUlFzjTLUyp8C3zvBbUGPF5Ss7ExTGeiDuMON6hRir2KcJbEX85ingisN9MyYZoIvYC1EgQmHFWhZ4x9PEEnhrK+GI3jbRxJcwOJZb4MLoooiuF0XCFpAhhpTTQ7VTKt3idoh5SopSfa1Qotc1xq40LBOOP0r2OcszYEpkLgBxkFwdNFBwjPYS6YTw52Ksqo1Zh0sIY4cZRfHcWJy1rOfKKBRroLzeI37y8hdKl0hfWgsqyLRSF8Ay4w1159ELVt64jX8CzhODcLptIwMtlOCZs42VBqtH2GBfKtzWyZAf5b377KcATO1hKWCr6KGPGnQUE+1iLw/gT8F9zUUbEcpzbF8O5glVsqtpnu7SQuW7gbrlANfSKpV9MmDO6T4UhwugJEJJ1bLyFG4OCYRXX+kS0Mu6m3eqY6MPzE9zk5Y1SoPfKflX5APt1nWkYxMLqNR8r2Z56p9XiZV+KgEp/Y7GltxK6Cuo4twoffFv6DXyL6HqMr5Woq9ou55JPCsg95QL0cZNpp6Qs2Tijm1XOtKLaIozTjhMyxYI0OOOuJVGT3RuXkTvgkvo4KtIzIUMZ7hYWqGNOYZlj/C3kFvfgkvfwt/t6zNVUtaGlZDeor5idKiFVXxhm3PkmI6hBUo/6tMfU+TkxSF2P9rz0z79y6mOaa7Z84f1cF5/b9ESl+Do7B0IreonfjD/Uc80HsvLEVWF4R+RXiHShS18dHB3tU8I7dHOOvtP/aOGd8+qA8HjVx1QRlHbTYs3qb5/SuhD+5ehVejLA4TmVLYz9UMN4yfTW9bTePhSyNoNgykDkZ+ECjmeLAwN5f6VbAaUwd+a92fAY3W1bVz16lM5srKGAs+d70zAz+pqgLNy3DU9vYXM8w6lqH1O0eDNiWUjUVrddKkTVLUGG6Yme1BYEUDCFYrGiJdB7c3RUuQmDju/zKQBQ1aSff1AjN58ow56u3XbdES6gzQCnJIa/7kou0GY2fDz5GqxwfaNA/3IHYGI2lZS/7yjqCCuLAD52zh/mBh0BIal8a0IEhofudF8XTtP18gFjCqEqL6KXUy4hkqxVJ0W+OSGpLax338Gnu91nezSSjKJz290zfAdCsR

TS error in App.vue

### Expected behavior

Should be able to detect common state (bananas in reproduction)

### Actual behavior

Unable to detect common state even though we are able to use it in the application

### Additional information

_No response_

jh-leong:
close #2785

### DIFF
--- a/packages/pinia/src/storeToRefs.ts
+++ b/packages/pinia/src/storeToRefs.ts
@@ -68,9 +68,11 @@ type _ToStateRefs<SS> =
  * Extracts the return type for `storeToRefs`.
  * Will convert any `getters` into `ComputedRef`.
  */
-export type StoreToRefs<SS extends StoreGeneric> = _ToStateRefs<SS> &
-  ToRefs<PiniaCustomStateProperties<StoreState<SS>>> &
-  _ToComputedRefs<StoreGetters<SS>>
+export type StoreToRefs<SS extends StoreGeneric> = SS extends unknown
+  ? _ToStateRefs<SS> &
+      ToRefs<PiniaCustomStateProperties<StoreState<SS>>> &
+      _ToComputedRefs<StoreGetters<SS>>
+  : never
 
 /**
  * Creates an object of references with all the state, getters, and plugin-added
--- a/packages/pinia/test-dts/store.test-d.ts
+++ b/packages/pinia/test-dts/store.test-d.ts
@@ -5,7 +5,7 @@ import {
   expectType,
   storeToRefs,
 } from './'
-import { computed, ref, UnwrapRef, watch } from 'vue'
+import { computed, Ref, ref, UnwrapRef, watch, WritableComputedRef } from 'vue'
 
 const useStore = defineStore({
   id: 'name',
@@ -328,3 +328,12 @@ expectType<number>(refs.bananasAmount.value)
 refs.bananasAmount.value = 0
 // @ts-expect-error: this one is readonly
 refs.total.value = 0
+
+const refStore = defineStore('ref-bananas', () => {
+  const bananas = ref(['banana1', 'banana2'])
+  return { bananas }
+})()
+declare const conditionalStore: typeof refStore | typeof writableComputedStore
+expectType<Ref<string[]> | WritableComputedRef<'banana'[]>>(
+  storeToRefs(conditionalStore).bananas
+)


packages/pinia/src/storeToRefs.ts
```diff
@@ -68,9 +68,11 @@ type _ToStateRefs<SS> =
  * Extracts the return type for `storeToRefs`.
  * Will convert any `getters` into `ComputedRef`.
  */
-export type StoreToRefs<SS extends StoreGeneric> = _ToStateRefs<SS> &
-  ToRefs<PiniaCustomStateProperties<StoreState<SS>>> &
-  _ToComputedRefs<StoreGetters<SS>>
+export type StoreToRefs<SS extends StoreGeneric> = SS extends unknown
```
posva, line 71: Why does this fix the problem? Isn't `x extends unknown` always true? I want to add a note comment for my future self and other contributors
TS is sometimes so weird 🫠

jh-leong: When conditional types act on a generic type, they become distributive when given a union type ([Distributive Conditional Types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types)).

In this case, it will always resolve to the first branch because `unknown` is the top type in TypeScript.

posva: Thanks! This makes me wonder if there isn't another bug up the chain of types. But this is already good enough!

posva (approved):
Thanks a lot!

PR #2795, opened 2024-10-08
fix: reference the store directly in storeToRefs to ensure correct reactivity after HMR

### LINKED ISSUE #2779: define var storeToRefs(store) HMR not work as expected
### Reproduction

https://stackblitz.com/edit/github-k5deut?file=src%2Fstores%2Fcart.ts,src%2FApp.vue,src%2Fstores%2Fuser.ts

Open src/stores/cart.ts
Change 
```
 test(state) {
    return state.totalNum + 1005;
  }
```
Check log
image

```App.vue
    const cart = useCartStore();
    const { test } = storeToRefs(cart);
    const testStoreToRefs = storeToRefs(cart)
    watchEffect(() => {
      console.log(
        cart.test,
        test.value,
        testStoreToRefs.test.value,
        storeToRefs(cart).test.value,
        storeToRefs(cart).test
      );
    });
```

### Expected behavior

when use 
```
    const { test } = storeToRefs(cart);
    const testStoreToRefs = storeToRefs(cart)
```
change file src/stores/cart.ts,   HMR should work as  cart.test(or storeToRefs(cart).test.value or storeToRefs(cart).test)

### Actual behavior

when use 
```
    const { test } = storeToRefs(cart);
    const testStoreToRefs = storeToRefs(cart)
```
change file src/stores/cart.ts,   HMR not work as expected

### Additional information

_No response_

babu-ch:
fix: https://github.com/vuejs/pinia/issues/2779

I have a store referenced with toRaw, so it doesn’t update when it changes. 
By referencing a store that is not wrapped in toRaw, it updates as expected. 
If it is necessary to reference something wrapped in toRaw, I apologize.

### DIFF
--- a/packages/pinia/__tests__/storeToRefs.spec.ts
+++ b/packages/pinia/__tests__/storeToRefs.spec.ts
@@ -1,6 +1,7 @@
 import { describe, beforeEach, it, expect } from 'vitest'
 import { computed, reactive, ref, ToRefs } from 'vue'
 import { createPinia, defineStore, setActivePinia, storeToRefs } from '../src'
+import { set } from 'vue-demi'
 
 describe('storeToRefs', () => {
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

PR #2796, opened 2024-10-08
Update actions type in mockStore helper method

Kapcash:
I recently tried the `mockedStore` helper function to have rich types on the mocked actions of a store, but it **failed** to compile.

I managed to use the correct type for **Vitest** and tested it's working correctly :)


posva:
Thanks! I adapted it in 67ddc04

PR #2798, opened 2024-10-13
feat(nuxt): do not serialize skipHydrate properties

### LINKED ISSUE #2773: State marked with `skipHydrate` should not be included in the Nuxt payload

I expected that `skipHydrate` would also leave the ref out of the Nuxt payload. Nuxt throws errors when it cannot serialize certain objects. I could keep unserializable stuff out of the pinia stores, but I don't like the idea of introducing multiple solutions for shared state. I would much prefer to use Pinia for all shared state.

[Example CodeSandbox](https://codesandbox.io/p/devbox/relaxed-smoke-d36z4f?file=%2Fstores%2FuseFooStore.ts)

### Proposed solution

Make `skipHydrate` also skip transferring of values to the nuxt payload.

_No response_

kabalage:
Fixes #2773
Let me know if anything else is needed.


posva (approved):
Thanks!

PR #2828, opened 2024-11-09
feat(nuxt): supports automatic import of stores from extend layers

wzc520pyfm:
When I install pinia in a nuxt ***layer*** and create `stores` folder, and then I extend this ***layer*** in another nuxt application, I find that the `stores` are not automatically imported.

### What this PR does:

When addImportsDir, consider the nuxt layers also.

#### Refer:

[Pinia autoImports for nuxt](https://pinia.vuejs.org/ssr/nuxt.html#Auto-imports)

[Nuxt layers](https://nuxt.com/docs/getting-started/layers)

### DIFF
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -101,6 +101,10 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
     if (options.storesDirs) {
       for (const storeDir of options.storesDirs) {
         addImportsDir(resolve(nuxt.options.rootDir, storeDir))
+
+        for (const layer of nuxt.options._layers) {
+          addImportsDir(resolve(layer.config.rootDir, storeDir))
+        }
       }
     }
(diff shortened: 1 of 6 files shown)


packages/nuxt/src/module.ts
```diff
@@ -101,6 +101,10 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
     if (options.storesDirs) {
       for (const storeDir of options.storesDirs) {
         addImportsDir(resolve(nuxt.options.rootDir, storeDir))
+
+        for (const layer of nuxt.options._layers) {
```
posva, line 105: Isn't this an internal property? 😓 Let's use a public API

posva (changes requested):

posva:
I noticed there is also https://github.com/vuejs/pinia/pull/2757, so let's use that one instead. Thanks for the PR!

PR #2847, opened 2024-11-28
feat: writable `computed`s to be picked up by `mapWritableState`

### LINKED ISSUE #2823: Allow writable `computed`s to be picked up by `mapWritableState`

At the moment, when attempting to use a writable `computed` (`computed({ get () { … }, set (value) { … } })`) with `mapWritableState`, TS reports an error. I assume that this is because `mapWritableState`, unlike `mapState`, only allows state and not `computed`s.

Despite TypeScript complaining, it seems like it does work as expected at runtime (but I did not test extensively).

https://play.pinia.vuejs.org/#eNqNVO9P2zAQ/VeOaFJSKXVATEyqYGKbmMQ+MDSm7cMyiZBci8GxI9tJO5X87zs7Pxo6hNZP9fm9u3vvct4GH6qKNTUGi+DU5JpXFkQmV2dpYE0avE8lLyulLWyhNnhjlUZoYalVCSFLclVLi5pZE06AZVb91Nxmd4IImd0RKi55RshU4sZjC1xmtSBOKgFyVVa1xWLRHQEYY/upoqGJGH6FMowhzDMplb3m+WP4e+Z4bZzKNpWnSaeGFNDBYlkJ4tMJ4JRLKgT2T4VOJm5sGkAzL1WBggK7jKQftluQ0PqEkyRBHHRy59QhezBKkn++7bS/IPNGIWnglbtIGiT+P0NTzu+0Whuy74HA1HUHPadhJAU2Vilh5lnFO9q9tZVZJEleSMJTp7zRTKJNZFUm/3DOT9gJO0oEv0uoUMJlgZvnZYgxL7D8n+wD9PyQHb1l73zW5rhPWrqs3neynWyxJldyyVd7prjhcoH6a2W5ks/NyYRQ6y8+ZnWNY4v5PeaPL8QfzKbr+loj2dfgRJbN9Appnu764ubKz3a8pAnXgtCvXH5Do0TteuxgH2tZUNsTnO/20s+Yy9V3c7GxKM0gyjU6fIW9zZ9ekb5r95gdT1zcLRZZOC4WrQuXe0s47NQIGtYoBo3LGNaZze9HNLUz3T8albG7xT6bVojCvgnasmgGZ7QK3ZY6iiQspY8OaeVc1FeJOphkTSZoXAPLU5RAJtQq6i9n0PbMLt9u5yjxoCDqbSKHXK7+BFTY1lrC7Zttn6297a460wGMI/ibBRiraUwTdi8A1z8cgMpVmTZ4KW3H8G+I+/ElRAcj7OkJuLnKrqIhMnOhA24+k//0Lo3hSSUYrKAih0OwBRQGXwYNSUZsL8t/GL1fvXh6leKpa+67IUTQ/gUzFOyv

### Proposed solution

Allow `computed`s that are writable (`WritableComputedRef`) to be used within `mapWritableState`.

_No response_

noootwo:
resolve #2823

### DIFF
--- a/packages/pinia/__tests__/mapHelpers.spec.ts
+++ b/packages/pinia/__tests__/mapHelpers.spec.ts
@@ -248,30 +248,23 @@ describe('Map Helpers', () => {
 
     it('setup store', async () => {
       const useSetupStore = defineStore('setup', () => {
-        const a = ref(true)
-        const n = ref(0)
+        const text = ref('initial')
 
-        const double = computed({
-          get: () => n.value + n.value,
+        const textUpper = computed({
+          get: () => text.value.toUpperCase(),
           set: (v) => {
-            n.value = v
-          },
-        })
-        const notA = computed({
-          get: () => !a.value,
-          set: (v) => {
-            a.value = v
+            text.value = v
           },
         })
 
-        return { a, n, double, notA }
+        return { text, textUpper }
       })
 
       await testComponent(
-        mapWritableState(useSetupStore, ['n', 'a', 'double', 'notA']),
-        `{{ n }} {{ a }} {{ double }} {{ notA }}`,
-        `0 true 0 false`,
-        'replaced replaced replacedreplaced false'
+        mapWritableState(useSetupStore, ['text', 'textUpper']),
+        `{{ text }} {{ textUpper }}`,
+        `initial INITIAL`,
+        'replaced REPLACED'
       )
     })
   })
--- a/packages/pinia/__tests__/mapHelpers.spec.ts
+++ b/packages/pinia/__tests__/mapHelpers.spec.ts
@@ -10,7 +10,7 @@ import {
   setMapStoreSuffix,
 } from '../src'
 import { mount } from '@vue/test-utils'
-import { nextTick, defineComponent } from 'vue'
+import { nextTick, defineComponent, ref, computed } from 'vue'
 import { mockWarn } from './vitest-mock-warn'
 
 describe('Map Helpers', () => {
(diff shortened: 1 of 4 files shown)


posva:
Nice!

packages/pinia/src/types.ts
```diff
@@ -572,7 +572,11 @@ export type _ActionsTree = Record<string, _Method>
  * For internal use **only**
  */
 export type _ExtractStateFromSetupStore_Keys<SS> = keyof {
```
posva: This type shouldn't include computed (even writable ones). The types related to `mapWritableState` should be the ones changed so they include writable computed alongside state properties

posva (changes requested):
Thanks!

packages/pinia/src/types.ts
noootwo: Ok, I get it

packages/pinia/test-dts/mapHelpers.test-d.ts
```diff
@@ -161,6 +165,17 @@ describe('mapHelpers', () => {
...
+      expectTypeOf<{
+        a: {
+          get: () => 'on' | 'off'
+          set: (v: 'on' | 'off') => any
+        }
+        writableUpper: {
+          get: () => string
+          set: (v: 'on' | 'off') => any
+        }
+      }>(mapWritableState(useSetupStore, ['a', 'writableUpper']))
```
posva: we can remove the `a`
```suggestion
      expectTypeOf<{
        writableUpper: {
          get: () => string
          set: (v: 'on' | 'off') => any
        }
      }>(mapWritableState(useSetupStore, ['writableUpper']))
```

packages/pinia/__tests__/mapHelpers.spec.ts
```diff
@@ -245,5 +245,34 @@ describe('Map Helpers', () => {
         'replaced replaced'
       )
     })
+
+    it('setup store', async () => {
+      const useSetupStore = defineStore('setup', () => {
+        const a = ref(true)
```
posva: Let's just have one single variable named `text` that starts as `'initial'` and the computed is `textUpper` which returns an uppercase value.

posva (changes requested):

posva (approved):
Thanks!

PR #2849, opened 2024-12-05
chore(playground): update dependencies and init

adamdehaven:
Updates the Pinia Online Playground

- Fetches and utilizes the latest version of Vue
- Fetches and utilizes the latest version of TypeScript
- Updates the to the latest version of `@vue/repl` (I'm not super-familiar, so check my updates here)
- Updates the README and associated package commands

I was going to update the `tsconfig.json` to utilize `"strict": true` but didn't dig in to where this is sourced.

### DIFF
--- a/packages/online-playground/package.json
+++ b/packages/online-playground/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@pinia/playground",
+  "name": "@pinia/online-playground",
   "version": "0.0.0",
   "type": "module",
   "private": true,
@@ -11,10 +11,11 @@
   "devDependencies": {
     "@vitejs/plugin-vue": "^5.2.1",
     "execa": "^9.5.1",
+    "typescript": "^5.7.2",
     "vite": "^6.0.2"
   },
   "dependencies": {
-    "@vue/repl": "^3.0.0",
+    "@vue/repl": "^4.4.2",
     "file-saver": "^2.0.5",
     "jszip": "^3.10.1",
     "pinia": "workspace:*",
--- a/packages/online-playground/src/App.vue
+++ b/packages/online-playground/src/App.vue
@@ -1,8 +1,8 @@
 <script setup lang="ts">
 import Header from './Header.vue'
-import { Repl, ReplStore, SFCOptions, ReplProps } from '@vue/repl'
+import { Repl, useStore, useVueImportMap, SFCOptions, ReplProps } from '@vue/repl'
 import Monaco from '@vue/repl/monaco-editor'
-import { ref, watchEffect, onMounted, provide } from 'vue'
+import { computed, ref, watchEffect, onMounted, provide } from 'vue'
 import { AppVue, PiniaVersionKey, counterTs } from './defaults'
 
 const setVH = () => {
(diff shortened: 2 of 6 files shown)


packages/online-playground/src/App.vue
```diff
@@ -62,25 +94,6 @@ provide(PiniaVersionKey, piniaVersion)
...
-            'vue-demi':
-              'https://cdn.jsdelivr.net/npm/vue-demi@0.14.7/lib/v3/index.mjs',
-          }
-        : {
-            pinia: '/src/pinia-dev-proxy',
-            vue: '/src/vue-dev-proxy',
-            'vue/server-renderer': '/src/vue-server-renderer-dev-proxy',
-          }),
-    },
-  })
```
adamdehaven, line 82: I didn't totally understand what this block was doing, and removing didn't seem to impact the playground...?

packages/online-playground/src/App.vue
```diff
@@ -90,12 +103,10 @@ if (!hash) {
 }
 
 // persist state
-watchEffect(() => {
-  const newHash = store
-    .serialize()
-    .replace(/^#/, useDevMode.value ? `#__DEV__` : `#`)
```
adamdehaven, line 96: I removed this `replace` logic; things still appear to be functional but please double-check

packages/online-playground/package.json
```diff
@@ -11,10 +11,11 @@
   "devDependencies": {
     "@vitejs/plugin-vue": "^5.2.1",
     "execa": "^9.5.1",
+    "typescript": "^5.7.2",
```
posva, line 14: Do we need this? ts is installed at the root to make sure only one version is used (currently 5.6 because other things break)

packages/online-playground/src/App.vue
```diff
@@ -19,15 +19,47 @@ if (hash.startsWith('__DEV__')) {
...
-  defaultVueRuntimeURL:
-    'https://cdn.jsdelivr.net/npm/@vue/runtime-dom@3.4.21/dist/runtime-dom.esm-browser.js',
-  defaultVueServerRendererURL:
-    'https://cdn.jsdelivr.net/npm/@vue/server-renderer@3.4.21/dist/server-renderer.esm-browser.js',
+const {
+  importMap: builtinImportMap,
+  vueVersion,
+  productionMode,
+} = useVueImportMap({
+  runtimeDev: 'https://cdn.jsdelivr.net/npm/vue@3/dist/vue.runtime.esm-browser.js',
```
posva, line 27: We actually really need to fetch the latest version from npm and use that specific tag. We cannot use `@3` as a default because the idea is that a link should always work or fail the same way

packages/online-playground/src/App.vue
```diff
@@ -19,15 +19,47 @@ if (hash.startsWith('__DEV__')) {
...
+            'vue-demi': 'https://cdn.jsdelivr.net/npm/vue-demi@0/lib/v3/index.mjs',
+            // Latest 5.x version
+            'typescript': 'https://cdn.jsdelivr.net/npm/typescript@5/lib/typescript.min.js',
+          }
+        : {
+            pinia: '/src/pinia-dev-proxy',
+            vue: '/src/vue-dev-proxy',
+            'vue/server-renderer': '/src/vue-server-renderer-dev-proxy',
+            // Latest 5.x version
+            'typescript': 'https://cdn.jsdelivr.net/npm/typescript@5/lib/typescript.min.js',
```
posva, line 52: We shouldn't need typescript, the [vue js playground doesn't have it](https://play.vuejs.org)

packages/online-playground/src/App.vue
posva: I think it's not needed anymore since it's done in the initialization

packages/online-playground/src/App.vue
posva: Maybe it's now the `__PROD__` that needs to be removed, maybe it's done by the serialize function now. For this kind of things I would look at the vuejs/core playground itself

packages/online-playground/src/App.vue
```diff
@@ -90,12 +103,10 @@ if (!hash) {
...
-watchEffect(() => {
-  const newHash = store
-    .serialize()
-    .replace(/^#/, useDevMode.value ? `#__DEV__` : `#`)
-  history.replaceState({}, '', newHash)
-})
+watchEffect(() => history.replaceState({}, '', store.serialize()))
+
+// production mode is enabled
+productionMode.value = import.meta.env.PROD
```
posva, line 109: We probably need to remove this. It's better to use the dev version by default (like in the vuejs playground)

posva (changes requested):
Thanks a lot!

- The `tsconfig` disappeared. We probably need to manually add the file (https://github.com/vuejs/repl/commit/ca548b240addd4f58851a545622684a00eb09a0a)
- The versions must be fully specified so I think we might need to fetch the latest version from npm when no hash is specified. It's fine to use Suspense for this

Issue #2848, opened 2024-12-03
However, when I pass a null value for a certain key into storeToRefs, it will throw an error.


Jv-Juven:
### Reproduction

-

### Steps to reproduce the bug

In the function definition within the file "pinia/packages/pinia/src/storeToRefs.ts", if the property value of the object rawStore obtained from `const rawStore = toRaw(store)` might be null, an error will occur.
![image](https://github.com/user-attachments/assets/822d5cfc-d335-47ed-972b-b79f44b20fb1)
![image](https://github.com/user-attachments/assets/91983964-3e6b-431c-90f9-abe024112550)

### Expected behavior

When the final computed value of rawStore is `{ projectId: null, name: null }`, the program can execute normally.

### Actual behavior

![image](https://github.com/user-attachments/assets/47467d56-3c96-4831-8132-0dc1618f5d34)

posva:
This simply sounds like wrong usage but you didn’t provide a repro like asked… open a discussion for help or use the discord, please

posva:
To help others who might have this issue. This comes from code like this:


```ts
export const useStore = defineStore('counter', () => {
  let thing = null

  return { thing }
})
```

This should be changed to

```ts
export const useStore = defineStore('counter', () => {
  let thing = ref(null)

  return { thing }
})
```

Otherwise, **things don't work**: the `thing` variable used inside of the store is different from the one exposed through `useStore().thing`. This is worth adding as an eslint rule.

---

Another scenario that wasn't meant to work (and actually breaks with some plugins) is this:

```ts
interface UserStore {
  userList: string[]
  user?: string
}

export const useUserStore = defineStore('user', {
  state: (): UserStore => ({
      userList: [],
      // user is omitted, needs to be `user: undefined` after this commit, to be able to use storeToRefs
  }),
})
```

The correct version would be:

```ts
interface UserStore {
  userList: string[]
  user: string | undefined // or `null`
}

export const useUserStore = defineStore('user', {
  state: (): UserStore => ({
      userList: [],
      user: undefined
  }),
})
```

PR #2854, opened 2024-12-07
feat: add null safe for store to refs

### LINKED ISSUE #2853: TypeError: Cannot read properties of undefined (reading 'effect')
### Reproduction

https://play.pinia.vuejs.org/#eNqNVN9v2jAQ/lesvBAmSFZ16iRGULeKh07aD7V9zIuJD+rOsS3boUyU/31nm4RAWVV4IL777rvvPnPZJl+1ztYNJJNkaivDtSMWXKOJoHJVlImzZTIrJa+1Mo5sSWPh3ikDZEeWRtVkkOWVaqQDkzk76AGtRz2oO1jaDqu55BRBpayUtNgpMBUdaTpsM1vCYNGsbpRc8hXWF32+NDwPv0SwEpAJtUp7BUgzzeM0qB0PDmotqAM8ETLVBmbbkw67ae7DvrAHTkZJHGhcU509WSXRp60nKfcJtGdCQsTHwoA+UiZ5eM7A1uOFUc8WDXpC8KiFXqPpOYO1U0rYMdU8lj06p+0kzysmEc9A8LXJJLhc6jp/VXN9lV1lF7ngixwb5Vwy2By3wYoxg/o97C30+mN28Sn7HFjXl3vS2rN60l0pd2iLs+g9OndiSqVqzQWYX9pxvJsjc6gQ6vl7iDnTQCexeoTqz5n4k91E1b8NoH1r6I3lqFmBi+n5/U/Y4HOXrBVrBKLfSN4B/m8arzHCvjWSoeweLqi9DXfM5erBzjcOpG2H8kKDGwEfbL55Y/SD3MvssufiYXXQwm51GCy5PFmzdnU6kIFll8TuYatKCZuQjkvU7WrRp0x9/wHVehC0p0NSzFqlsa6/GIVvlO7ThNhG+wYTsqTCtldFCEi6EMBOw+pZgtkHY2znt9Y/CHAEqkdq3C22pLKCyTwcbXsmL0Q2QqAC/9Mrq5WklZozjtN0xSSeH/5qyG7vMcaoUBIvhUHMtHQvBK86uMEO1P57mB9fgFELsiMmHRazzoAT0ZgmH/Czn+4Vz48zWlvKouM8N9F/mA2+nI0k294djY4ljc6xjfoz+cM5Ydgk3lGy+wcHkg9N

I need to store some data in the store, but it is not ref type, also need to initialize null, but now the error, this is a big destructive update, I think you should set to? .effect, like some echarts or monaco editor examples, is not a ref and has to wait for some work to be done, so it starts out null or undefined, which I think the plugin should allow

### Expected behavior
image

### Actual behavior

I need to store some data in the store, but it is not ref type, also need to initialize null, but now the error, this is a big destructive update, I think you should set to? .effect, like some echarts or monaco editor examples, is not a ref and has to wait for some work to be done, so it starts out null or undefined, which I think the plugin should allow

### Additional information

I think it should be compatible with previous behavior, ref type or  change to ? .effect  https://github.com/vuejs/pinia/commit/67d31094784cc3bd256b0636b79dc8e421f6c3fb#diff-a4019455d4214f4ae04e5e463a0b67af62b68c0267b436ab16beeaaf45bc32d3R105

jahnli:
close https://github.com/vuejs/pinia/issues/2853

Should be compatible with the old script, allowing initialization to null as well as non-ref scenarios

### DIFF
--- a/packages/pinia/src/storeToRefs.ts
+++ b/packages/pinia/src/storeToRefs.ts
@@ -102,7 +102,7 @@ export function storeToRefs<SS extends StoreGeneric>(
       const value = rawStore[key]
       // There is no native method to check for a computed
       // https://github.com/vuejs/core/pull/4165
-      if (value.effect) {
+      if (value?.effect) {
         // @ts-expect-error: too hard to type correctly
         refs[key] =
           // ...


posva:
Thank you for the PR. See https://github.com/vuejs/pinia/issues/2848#issuecomment-2516779166 for anti patterns that lead to this error and how to fix them. If you found a different approach that also creates the problem, please let me know so I can add them.

If you have the error, it's important to **adapt it** as it means your code won't work correctly as shown in the issue

jahnli:
> 感谢您的 PR。请参阅[#2848（评论）](https://github.com/vuejs/pinia/issues/2848#issuecomment-2516779166)以了解导致此错误的反模式以及如何修复它们。如果您发现其他方法也会导致此问题，请告诉我，以便我可以添加它们。
> 
> 如果你遇到错误，请务必**进行调整**，因为这意味着你的代码将无法正常工作，如问题所示
image



> Thank you for the PR. See [#2848 (comment)](https://github.com/vuejs/pinia/issues/2848#issuecomment-2516779166) for anti patterns that lead to this error and how to fix them. If you found a different approach that also creates the problem, please let me know so I can add them.
> 
> If you have the error, it's important to **adapt it** as it means your code won't work correctly as shown in the issue

I see the need to set something like echartsInstance to ref(null), but actually these are not reactive data, is there a better solution for this scenario,

1. No ref response is required
2. The initial value must be set to null

posva:
shallowRef or not returning them

PR #2857, opened 2024-12-11
feat: throw when a store is used outside of a Nuxt-aware context.

pguilbert:
## Context
The Pinia documentation states the following:
> Make sure all of your `useStore()` calls appear before any `await`. Otherwise, this could lead to using the wrong Pinia instance in SSR apps: [pinia.vuejs.org/cookbook/composing-stores.html#Shared-Actions](https://pinia.vuejs.org/cookbook/composing-stores.html#Shared-Actions)

We recently discovered that we had cross-request pollution caused by bad use of Pinia stores after an await in Nuxt. With a large codebase, it was tough to track down every instance of the issue. So, we ended up creating a custom version of `defineStore`, similar to the one in this PR, to help prevent developers from misusing Pinia stores.

I think it would be useful if this could be implemented directly in the Pinia Nuxt module.

## Description of the changes

The [Nuxt Pinia module](https://github.com/vuejs/pinia/blob/v2/packages/nuxt/src/runtime/plugin.vue3.ts) does the following for each request:

```js
// Create the instance for this request
const pinia = createPinia()
// Store the instance within the Nuxt context (mainly for serialization after)
nuxtApp.vueApp.use(pinia)
// Set the Pinia global instance (that will be used by Pinia and is subject to cross-request pollution)
setActivePinia(pinia)
```

The module also provides a helper `usePinia()` that retrieves the **Pinia instance from the Nuxt context and not from the Pinia global variable**. Conveniently, this helper will throw an error if used after an `await` (outside of a Nuxt-aware context)!

This PR ensures that on the server we always use `usePinia()` to get the current Pinia instance. By doing so `useStore` will systematically throw if used after an await on the server (similarly to the `useNuxtApp` composable) 

## Breaking change

Server-side, using a store outside a Nuxt-aware context will now result in an error. Previously, it would continue without error, but sometimes, the wrong Pinia instance was used.

Error in dev: 
> [nuxt] A composable that requires access to the Nuxt instance was called outside of a plugin, Nuxt hook, Nuxt middleware, or Vue setup function

Error in prod:
> [nuxt] instance unavailable

## Summary by CodeRabbit

* **New Features**
  * Stores now work reliably in server-side rendering environments with improved context awareness and error handling
  * Enhanced error messages when stores are accessed outside the proper Nuxt context

* **Documentation**
  * Added example demonstrating store usage patterns in server-side rendering scenarios

* **Tests**
  * Added test coverage for store behavior across various server-side rendering contexts

### DIFF
--- a/packages/nuxt/global.d.ts
+++ b/packages/nuxt/global.d.ts
@@ -1,0 +1,2 @@
+// Global compile-time constants
+declare var __TEST__: boolean
--- a/packages/nuxt/src/runtime/composables.ts
+++ b/packages/nuxt/src/runtime/composables.ts
@@ -1,5 +1,33 @@
 import { useNuxtApp } from '#app'
-import type { Pinia } from 'pinia'
+import {
+  defineStore as _defineStore,
+  type Pinia,
+  type StoreGeneric,
+} from 'pinia'
 export * from 'pinia'
 
-export const usePinia = () => useNuxtApp().$pinia as Pinia
+export const usePinia = () => useNuxtApp().$pinia as Pinia | undefined
+
+export const defineStore: typeof _defineStore =
+  process.env.NODE_ENV === 'production' && !__TEST__
+    ? _defineStore
+    : (((
+        ...args: Parameters<typeof _defineStore>
+      ): ReturnType<typeof _defineStore> => {
+        if (!import.meta.server) {
+          return _defineStore(...args)
+        }
+
+        const originalUseStore = _defineStore(...args)
+        function useStore(
+          pinia?: Pinia | null,
+          hot?: StoreGeneric
+        ): StoreGeneric {
+          return originalUseStore(pinia || usePinia(), hot)
+        }
+
+        useStore.$id = originalUseStore.$id
+        useStore._pinia = originalUseStore._pinia
+
+        return useStore
+      }) as typeof _defineStore)
--- a/packages/nuxt/test/nuxt.spec.ts
+++ b/packages/nuxt/test/nuxt.spec.ts
@@ -37,4 +37,8 @@ describe('Nuxt', async () => {
   it('can auto import from layers', async () => {
     expect(await $fetch('/')).toContain('Layer store: 0')
   })
+
+  it('throws an error server-side when the nuxt context is not available', async () => {
+    await expect($fetch('/usage-after-await')).rejects.toThrow()
+  })
 })
--- a/packages/nuxt/test/nuxt.spec.ts
+++ b/packages/nuxt/test/nuxt.spec.ts
@@ -39,6 +39,8 @@ describe('Nuxt', async () => {
   })
 
   it('throws an error server-side when the nuxt context is not available', async () => {
-    await expect($fetch('/usage-after-await')).rejects.toThrow()
+    await expect($fetch('/usage-after-await')).rejects.toThrowError(
+      '[nuxt] instance unavailable'
+    )
   })
 })
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@
 </p>
 <br/>
 <p align="center">
-  <a href="https://npmjs.com/package/pinia"><img src="https://badgen.net/npm/v/pinia" alt="npm package"></a>
-  <a href="https://github.com/vuejs/pinia/actions/workflows/ci.yml"><img src="https://github.com/vuejs/pinia/actions/workflows/ci.yml/badge.svg" alt="build status"></a>
-  <a href="https://codecov.io/gh/vuejs/pinia"><img src="https://codecov.io/gh/vuejs/pinia/graph/badge.svg?token=rU2xxQ6BGH"/></a>
+  <a href="https://npmjs.com/package/pinia"><img src="https://badgen.net/npm/v/pinia/v2" alt="npm package"></a>
+  <a href="https://github.com/vuejs/pinia/actions/workflows/ci.yml"><img src="https://github.com/vuejs/pinia/actions/workflows/ci.yml/badge.svg?branch=v2" alt="build status"></a>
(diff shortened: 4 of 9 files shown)


posva (changes requested):
Thanks for the PR! Do you have a reproduction of the problem where correctly using the `useStore()` composable returned by `defineStore()` doesn't work?

pguilbert:
> Thanks for the PR! Do you have a reproduction of the problem where correctly using the `useStore()` composable returned by `defineStore()` doesn't work?

"correctly" no. The purpose of this PR is to ease the detection of incorrect usage. In the current version bad usage of a store after a await will work without warning in low-traffic environments. But, in cases of simultaneous requests, there is a possibility that pinia uses the pinia instance from another request. 

I can provide a reproduction if needed, but this is already a documented case:
> Make sure all of your useStore() calls appear before any await. Otherwise, this could lead to using the wrong Pinia instance in SSR apps: [pinia.vuejs.org/cookbook/composing-stores.html#Shared-Actions](https://pinia.vuejs.org/cookbook/composing-stores.html#Shared-Actions)


For context: We recently found out that we had cross-request pollution caused by bad use of Pinia in Nuxt. With our large codebase, it was tough to track down every instance of the issue. So, we ended up creating a custom version of `defineStore`, similar to this one, to help prevent developers from misusing Pinia stores.

(I'll update the PR description with those infos.)

packages/nuxt/test/nuxt.spec.ts
```diff
@@ -33,4 +33,8 @@ describe('works with nuxt', async () => {
     expect(html).not.toContain('I should not be serialized or hydrated')
     expect(html).toContain('skipHydrate-wrapped state is correct')
   })
+
+  it('throws an error server-side when the nuxt context is not available', async () => {
+    await expect($fetch('/usage-after-await')).rejects.toThrow()
```
posva: Let's make this a bit stricter and include part of the error message with `toThrowError()` or similar

posva:
Thanks for the great description of the problem! I agree it's a nice addition

packages/nuxt/src/runtime/composables.ts
```diff
@@ -1,4 +1,27 @@
...
+
+  function useStore(pinia?: Pinia | null, hot?: StoreGeneric): StoreGeneric {
+    if (pinia) {
+      return store(pinia, hot)
+    }
+
+    return store(usePinia(), hot)
+  }
+
+  return useStore
```
posva:
```suggestion
  const originalUseStore = _defineStore(...args)

  return function useStore(pinia?: Pinia | null, hot?: StoreGeneric): StoreGeneric {
    return originalUseStore(pinia || usePinia(), hot)
  }
```

packages/nuxt/src/runtime/composables.ts
```diff
@@ -1,4 +1,27 @@
...
+import {
+  defineStore as _defineStore,
+  type Pinia,
+  type StoreGeneric,
+} from 'pinia'
 export * from 'pinia'
 
 export const usePinia = () => useNuxtApp().$pinia
+
+export const defineStore = (...args) => {
```
posva: Isn't this going to break the types? we can probably use `const defineStore: typeof _defineStore` or just cast the function as `as typeof `_defineStore`

posva (changes requested):

pguilbert:
> Thanks for the great description of the problem! I agree it's a nice addition

Nice! I'll complete the typing and refine the test to move this PR out of draft.

pguilbert:
@posva what do you think of this version? Could we merge this? 🙏

packages/nuxt/src/runtime/composables.ts
```diff
@@ -1,4 +1,27 @@
...
+  defineStore as _defineStore,
+  type Pinia,
+  type StoreGeneric,
+} from 'pinia'
 export * from 'pinia'
 
-export const usePinia = () => useNuxtApp().$pinia
+export const usePinia = () => useNuxtApp().$pinia as Pinia | undefined
+
+export const defineStore: typeof _defineStore = (
```
posva: Let's make the error dev only, it shouldn't affect the final bundle. This should be possible by using `process.env.NODE_ENV === 'production' ? _defineStore : ...` or a similar guard

posva (changes requested):

packages/nuxt/src/runtime/composables.ts
pguilbert: Sorry it took so long; it's done now!

iPrytz:
Thanks for this @pguilbert ! I think this is a great addition to Pinia. Is this planned to merge any time soon? 😊 @posva

packages/nuxt/global.d.ts
```diff
@@ -0,0 +1,2 @@
+// Global compile-time constants
+declare var __TEST__: boolean
```
posva, line 2: Any way to get rid of this? I don't think we should need this

packages/nuxt/src/runtime/composables.ts
```diff
@@ -1,5 +1,33 @@
...
-export const usePinia = () => useNuxtApp().$pinia as Pinia
+export const usePinia = () => useNuxtApp().$pinia as Pinia | undefined
+
+export const defineStore: typeof _defineStore =
+  process.env.NODE_ENV === 'production' && !__TEST__
+    ? _defineStore
+    : (((
+        ...args: Parameters<typeof _defineStore>
+      ): ReturnType<typeof _defineStore> => {
+        if (!import.meta.server) {
```
posva, line 17: This condition can be probably be moved up so we simply return `_defineStore` too

posva (changes requested):

PR #2858, opened 2024-12-13
fix(devtools): setup the settings panel immediately

### LINKED ISSUE #2818: Pinia devtools will not call `setupDevtoolsPlugin` until `useStore` is called
### Reproduction

N/A

There is a related issue https://github.com/vuejs/devtools/issues/672

The reproduction steps are there.

### Expected behavior

Call `setupDevtoolsPlugin` immediately instead of calling it after `useStore`. otherwise devtools cannot read `settings` at mounted.

### Actual behavior

Call `setupDevtoolsPlugin` after someone calls `useStore`

### Additional information

_No response_

noootwo:
fix #2818


posva:
Is this still needed in the new version?

noootwo:
> Is this still needed in the new version?

@posva Yes, the option about `settings` should be set when register pinia devtool.

posva:
I see. I think the fix should be different, registering the plugin once would be better. I will see later on if I can refactor the code for it

posva:
I tried fixing this with https://github.com/vuejs/pinia/tree/fix/devtools-setup-once but it seems like the timeline doesn't show up anymore and there are build issues. I don't know if this is a vue devtools regression

Issue #610, opened 2021-08-12
Speed could be better because of subscribe


posva:
### Reproduction

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Perf regression</title>
  </head>
  <body>
    <script src="https://unpkg.com/vue@3.2.2/dist/vue.global.js"></script>
    <script src="https://unpkg.com/vue-demi@0.11.2/lib/index.iife.js"></script>
    <script src="https://unpkg.com/pinia@2.0.0-rc.4/dist/pinia.global.js"></script>
    <!-- <script src="https://unpkg.com/pinia@2.0.0-beta.5/dist/pinia.global.js"></script> -->

    <div id="app">
      <section>
        <h1>Hello {{ store.$id }}</h1>

        <button @click="changeRandom">Change</button>

        <table class="center">
          <tr v-for="item in store.items" :key="item.id">
            <td>{{item.id}}</td>
            <td>{{item.name}}</td>
            <td>{{item.description}}</td>
          </tr>
        </table>
      </section>
    </div>

    <script>
      const {
        createApp,
        ref,
        reactive,
        toRaw,
        toRef,
        toRefs,
        onMounted,
        computed,
        effectScope,
      } = Vue

      const { defineStore, createPinia } = Pinia

      function makeRandomText(length) {
        let result = ''
        const characters =
          'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789'
        const charactersLength = characters.length
        for (let i = 0; i < length; i += 1) {
          result += characters.charAt(
            Math.floor(Math.random() * charactersLength)
          )
        }
        return result
      }

      const getName = () => makeRandomText(Math.ceil(Math.random() * 10) + 5)
      const getDescription = () =>
        makeRandomText(Math.ceil(Math.random() * 25) + 15)

      const useItemStore = defineStore({
        id: 'items',
        state: () => ({
          items: [],
        }),
        actions: {
          setItems(items) {
            console.time('setItems')
            this.items = [...items]
            console.timeEnd('setItems')
          },
          updateItem(item) {
            const index = this.items.findIndex((i) => i.id === item.id)
            console.time('updateItem')
            this.items.splice(index, 1, { ...item })
            console.timeEnd('updateItem')
[...]

PR #2870, opened 2025-01-04
fix: async subscription do not run when mutate state synchronously after patch

### LINKED ISSUE #992: `$subscribe`  miss mutations with type of `direct`  immediately after`patch`  mutations
in $subscribe ,I use `console` to print the mutations and states when the states  change : 
```javascript
 store.$subscribe((mutations, state) => {
    console.log('states change :', mutations, state);
  });
```

it works correct when I change the state directly or  use  `$pacth`  singlely
```javascript
// the both ways of the follow work correct
// 1. change directly : ouput is { type：‘direct’ ,... } 
 store.count++


// 2.use $patch : output is {type: 'patch object',...}
store.$patch({
    key: `${Math.random()}`,
 });
```

but when I use them together as follow,the mutation with type of 'direct' will miss ,there is only {type: 'patch object',...} on console , and the `count` value  printed was still the old value， actually had increased .  seem to not be detected ？
```javascript
store.$patch({
    key: `${Math.random()}`,
 });
store.count++
```

yangmingshan:
Fix #992

I run into same issue when I porting Pinia to Vue Mini, and I found another debugger issue for sync subscription (please see the test).

I managed a way to fix them, the fix may have extra memory / performance costs, I didn't do any measurements. But I used it in [Pinia for Vue Mini](https://github.com/vue-mini/vue-mini/blob/master/packages/pinia/src/store.ts#L297) anyway.

## Summary by CodeRabbit

* **Tests**
  * Updated expectations for subscription synchronization and added a test covering debugger events when a subscription is not triggered by a patch.

* **Improvements**
  * More consistent subscription triggering so events are delivered reliably after patches and direct synchronous mutations.

### DIFF
--- a/packages/pinia/__tests__/subscriptions.spec.ts
+++ b/packages/pinia/__tests__/subscriptions.spec.ts
@@ -368,12 +368,23 @@ describe('Subscriptions', () => {
       store.user = 'a'
       expect(syncSpy).toHaveBeenCalledTimes(2)
 
-      // FIXME: ideally, these should be 2 but we cannot use
-      // a sync flush within the store's $subscribe method
-      // https://github.com/vuejs/pinia/issues/610
       await nextTick()
-      expect(preSpy).toHaveBeenCalledTimes(1)
-      expect(postSpy).toHaveBeenCalledTimes(1)
+      expect(preSpy).toHaveBeenCalledTimes(2)
+      expect(postSpy).toHaveBeenCalledTimes(2)
+    })
+
+    it('debuggerEvents is not an array when subscription is not triggered by patch', () => {
+      const store = useStore()
+      store.$subscribe(
+        ({ type, events }) => {
+          if (type === MutationType.direct) {
+            expect(Array.isArray(events)).toBe(false)
+          }
+        },
+        { flush: 'sync' }
+      )
+      store.$patch({ user: 'Edu' })
+      store.user = 'a'
     })
   })
 })
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -249,7 +249,7 @@ function createSetupStore<
       if (isListening) {
         debuggerEvents = event
         // avoid triggering this while the store is being built and the state is being set in pinia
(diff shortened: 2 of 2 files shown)


posva:
Thanks! I will take a look when I can

yangmingshan:
This PR is now based on `v3` and can be merged

Does `v2` need this fix? If so, I can open another PR to fix `v2`

posva:
Thanks! I think it's fine to keep this for v3 only so don't worry about it

packages/pinia/src/store.ts
```diff
@@ -444,11 +434,19 @@ function createSetupStore<
...
       )
-      const stopWatcher = scope.run(() =>
-        watch(
+      const stopWatcher = scope.run(() => {
+        const stop1 = watch(
+          () => pinia.state.value[$id],
+          () => {
+            shouldTrigger = isListening
+          },
+          { deep: true, flush: 'sync' }
```
posva, line 443: I'm revisitting this PR and realizing that this might impact perf too much

posva: I need to see in the playground the impact

posva: Without checking the playground, I think we can't go this way: https://github.com/vuejs/pinia/issues/610

packages/pinia/__tests__/subscriptions.spec.ts
```diff
@@ -353,5 +353,35 @@ describe('Subscriptions', () => {
       expect(spy1).toHaveBeenCalledTimes(1)
       expect(spy2).toHaveBeenCalledTimes(2)
     })
+
+    it('debuggerEvents is not be an array when subscription is not trigger by patch', () => {
```
posva:
```suggestion
    it('debuggerEvents is not an array when subscription is not triggered by patch', () => {
```

posva:
Because of the perf degradation with `flush: 'sync'`, we can't go forward with this. I think it might be useful as a patch that can be applied with `pnpm` so it's usable for vue mini.

PR #2873, opened 2025-01-06
docs: remove duplicate entries in changelog

CamilleDrapier:
I just noticed there were some duplicated entries in the CHANGELOG, so I'm suggesting a simple fix here 🙇 

NB: I also noticed afterwards that the links/tags used for the version header were incorrect, so I'm also including a small fix for those, but I'm not sure if those changes are intended. Please let me know if this should be done differently or separately.


posva:
Oh sorry, I merged #2874 instead, I didn't see the order. Thanks anyway, this is always appreciated!

PR #2890, opened 2025-01-18
fix(nuxt): usePinia composable type

th1m0:
This PR fixes a type issue in the nuxt module. using the composable `usePinia` would always return `unknown` while it should always return the `Pinia` type.

Before: composable.d.ts

```ts
export * from 'pinia';
export declare const usePinia: () => unknown;
```

After: composable.d.ts

```ts
import type { Pinia } from 'pinia';
export * from 'pinia';
export declare const usePinia: () => Pinia;
```


th1m0:
@posva would be lovely if this could be merged as I currently have to always use `as Pinia` after `usePinia()`

```ts
	const pinia = usePinia() as Pinia // `as Pinia` is required here because it's Unknown
    const profileStore = useProfileStore(pinia)
```

posva:
Does this happen in a blank project? It's the first time I see this issue

th1m0:
Seems like it does, the type of the composable does not get into the build and defaults to `unknown`

https://stackblitz.com/edit/github-kd8twaqz?file=node_modules%2F%40pinia%2Fnuxt%2Fdist%2Fruntime%2Fcomposables.d.ts

posva:
It's so weird, this used to work so maybe a regression in Nuxt. I will merge but ~~release later on after doing a pass on other Nuxt related PRs. Aiming for the end of the month~~. It's released. Note you can use `getActivePinia()`, it does the same thing, maybe I should just deprecate `usePinia()`

posva (approved):

PR #2905, opened 2025-02-01
Pinia v3

posva:
- Remove Vue 2 support
  - Remove vue-demi
  - Remove workarounds for Vue 2
  - Remove Nuxt bridge (Nuxt 2) support
- Remove deprecated APIs
  - No more `defineStore({ id: 'id' })`
  - No `PiniaStorePlugin` (use `PiniaPlugin`)
- Upgrade `@vue/devtools-api` to v7
- Remove older TS support (anything `>=5` should be good)

TODO:

- ~~Write migration guide v2 to v3~~
- ~~Update docs to not mention Vue 2~~
- ~~Create a deployment for v2 docs~~
  - ~~Use local search only~~
- Migrate to `type: module` in package.json

It would be great to also ensure the Nuxt plugin works well even with layers so we can release a v1 instead of staying in 0.x when the nuxt module has been stable for a long time 😆


posva:
v3 is now the default branch

PR #2908, opened 2025-02-09
test: deprecated syntax

andrewvasilchuk:
Updated tests to avoid using the [deprecated syntax](https://pinia.vuejs.org/api/pinia/functions/defineStore.html#defineStore-id-storeSetup-options-) of the `defineStore` function.


posva:
Thanks! This was already added in the PR for pinia 3

PR #2911, opened 2025-02-12
feat: add madvue banner

IsraelOrtuno:
Add MadVue 2025 banner

### DIFF
--- a/packages/docs/.vitepress/theme/components/MadVueBanner.vue
+++ b/packages/docs/.vitepress/theme/components/MadVueBanner.vue
@@ -76,7 +76,7 @@
 
     </div>
     <a target="_blank" class="action" :href="target">
-      Know more
+      More Info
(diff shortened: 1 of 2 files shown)


packages/docs/.vitepress/theme/components/MadVueBanner.vue
```diff
@@ -0,0 +1,269 @@
...
+const closeBanner = () => {
+  // Hide the banner
+  isVisible.value = false
+  // Save action in the local storage
+  localStorage.setItem(nameStorage, String(true))
+  resetLayoutTopHeight()
+}
+
+onMounted(() => {
+  isVisible.value = !localStorage.getItem(nameStorage)
```
posva, line 115: Let's check all keys in localStorage that include `FREEWEEKEND25`. There must at least be one of them

```ts
const keys = Object.keys(localStorage).filter(key => key.includes('FREEWEEKEND25') $$ key.endsWith('_CLOSED'));

if (keys.length > 0 && keys.every(key => localStorage.getItem(key) != null)) {
    // ...
}
```

PR #2915, opened 2025-02-15
fix(nuxt): global pinia instance leads to whole nuxt context not being garbage collected

MirkoJa:
I discovered this during testing of a memory leak in our nuxt application. We were providing our apollo client in a plugin as well and it would never be garbage collected.

Relatively simple to test. 
- setup a nuxt project with pinia (not even a store necessary). 
- build the project
- run it with
```sh
node --inspect --expose-gc  .output/server/index.mjs
```
- use e.g. chrome to take a node heap snapshot
- with pinia the third object there will appear -> without pinia i doesn't appear as it is garbace collected

Reason:
The reason behind it is, that calling setActivePinia(pinia) writes it in a global variable, which is not cleared after ssr is done -> the same reference is passed to the nuxt plugin provide return. Keeping the activePinia instance leads to the whole nuxt context not being garbage collected.

grafik
In the Screenshot you can see that the activePinia keeps everything alive even after ssr is done.

Sadly I have no idea how to write a test for that.

(This is my first open source contribution so i might have no idea what i am doing :D)

### DIFF
--- a/packages/nuxt/src/runtime/plugin.vue3.ts
+++ b/packages/nuxt/src/runtime/plugin.vue3.ts
@@ -1,7 +1,6 @@
 import { createPinia, setActivePinia } from 'pinia'
 import type { Pinia } from 'pinia'
-import { defineNuxtPlugin, type Plugin } from '#app'
-import { toRaw } from 'vue'
+import { defineNuxtPlugin, useNuxtApp, type Plugin } from '#app'
 
 const plugin: Plugin<{ pinia: Pinia }> = defineNuxtPlugin({
   name: 'pinia',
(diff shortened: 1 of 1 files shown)


posva:
Nice find, thank you! In practice, this shouldn't matter since there is always 1 instance of pinia, so it won't create a memory leak, only holding of the memory. That being said, unsetting the global pinia instance on server after the app is sent seems reasonable!

MirkoJa:
The only thing i currently don't understand is if it makes any difference to link the reference to the nuxt payload when initializing the plugin. like this `nuxtApp.payload.pinia = toRaw(pinia.state.value)` vs. doing it in the `app:rendered` hook.

posva:
It shouldn't make a difference but doing it in `app:rendered` could be better.

posva (approved):

PR #2916, opened 2025-02-21
docs: update nuxt docs to recommend callOnce

TheAlexLichter:
This aligns well with https://nuxt.com/docs/getting-started/state-management#usage-with-pinia and recommends `callOnce` from Nuxt instead of `useAsyncData` (which should be side-effect free!)


posva (approved):
Thanks! Didn't know about the mode navigation

TheAlexLichter:
@posva It is rather new, we came up with that during the breakfast after Vue Toronto :D

PR #2923, opened 2025-02-24
docs: add devtools v3 compatibility note

### LINKED ISSUE #2919: pinia-devtools-plugin not working with Vue Devtools v6
### Reproduction

.

I see Pinia v3 dropped support for Vue 2, but I didn't see any mention of this dropping support for Devtools v6.
Devtools v6 support both Vue 2 and Vue 3.
Is this expected ?

I need Devtools v6 because I support both Vue 2 and Vue 3 project.
It's inconvenient to have to disable Devtools  v6 and enable Devtools v7 every time I switch project.

### Expected behavior

Maybe Pinia could keep support for both Devtools v6 and v7, but I don't know how much work that would be.
Otherwise, I hope Devtools v7 add support for Vue 2, but I wouldn't hold my breath as Vue 2 EOL.

### Actual behavior

.

### Additional information

_No response_

mrleblanc101:
fix #2919

### DIFF
--- a/packages/docs/cookbook/migration-v2-v3.md
+++ b/packages/docs/cookbook/migration-v2-v3.md
@@ -28,6 +28,9 @@ This deprecated type alias has been removed in favor of `PiniaPlugin`.
 - TypeScript 5 or newer is required.
 - The devtools API has been upgraded to [v7](https://devtools.vuejs.org).
 
+> [!NOTE]  
+> To use Pinia with Devtools v6, use Pinia 2
+
 ## Nuxt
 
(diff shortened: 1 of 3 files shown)


mrleblanc101:
@posva I'm a noob when it comes to this, I have no idea how to test.
I tried npm link, but couldn't figure it out exactly.
So I edited the `pinia.mjs` file directly in my browser and it seems to work.

Here is the output:
Screenshot 2025-02-24 at 7 51 24 PM

I confirm I see the message when using v6, and do not see the message when using v7.
The code for `isLegacyVueDevTools` comes from here: https://github.com/vuejs/devtools/blob/9462f0050c1dc06df56e0144b424acf26cbb5d3a/packages/devtools-kit/src/compat/index.ts#L4

PR #2939, opened 2025-03-15
Add production build for esm-browser

### LINKED ISSUE #2205: Bundle ESM for Vue 3 prod

### Proposed solution

porfirioribeiro:
Change the rollup config to also build pinia for esm browsers prod It will remove devtools and minify the code

Closes #2205

## Summary by CodeRabbit

* **New Features**
  * Introduced a minified production browser bundle for faster load times.
  * Added a distinct ESM browser output filename (esm-browser.prod) for easier identification and CDN consumption.

* **Chores**
  * Improved classification of browser builds as raw ESM to enhance compatibility with modern tooling.
  * Adjusted dependency handling for the browser production build to simplify integration.
  * No changes to public APIs.


porfirioribeiro:
I would appreciate it if you could take a look at this PR, @posva.

Will be really helpful to be able to use Pinia with Micro-Frontends using importmaps.

posva:
Hey @porfirioribeiro! Thanks for the PR, I still haven't got the time to take a look but I wanted to tell you that if you rebase from (or merge) `v3`, it should activate automatic releases so it can be used temporarily like these: https://github.com/vuejs/pinia/runs/47624782567

porfirioribeiro:
> Hey @porfirioribeiro! Thanks for the PR, I still haven't got the time to take a look but I wanted to tell you that if you rebase from (or merge) `v3`, it should activate automatic releases so it can be used temporarily like these: https://github.com/vuejs/pinia/runs/47624782567

Thanks @posva, whenever you have time take a look at this.

The temporarily deployed URLs will be helpful for testing. But would be nice to have this official so it could be part of the other npm CDNs

Thank you

skirtles-code (commented):
I did some local testing and these changes do seem to achieve the desired outcome. The `esm-browser.prod` appears to be generated correctly and the other output files remain unchanged. I have some minor suggestions below, but these are essentially just refactorings and I believe this PR could be merged in its present form.

---

I implemented something similar for Vue Router. See:

- https://github.com/vuejs/router/pull/2242/files

A key difference is that I used `moduleSideEffects: false` to ensure `@vue/devtools-api` was treeshaken, rather than removing it from the `external` dependencies. To be honest, I didn't actually consider this alternative approach. However, having given it a bit more thought I still think using `moduleSideEffects: false` is preferable. My reasoning here is:

1. It more explicitly states the desired intent. Conditionally removing `@vue/devtools-api` from `external` gives the impression that the dependency is sometimes inlined, which isn't really what's going on.
2. The approach used here needs `browser-prod` to be passed as the `buildName` to `createConfig`, essentially as a sentinel for 'magic mode'. The other builds don't have special handling for production builds in `createConfig` and ideally I think that should be maintained.
3. It's easier to check that it actually worked. With `moduleSideEffects: false` you just need to check that the built file doesn't import anything from `@vue/devtools-api`. With the inlining approach there isn't an easy way to confirm that nothing actually got inlined. The file size is a clue, but that doesn't necessarily rule out some small amount of unnecessary code being pulled in from `@vue/devtools-api`.

Personally, I would change to using `moduleSideEffects: false` and revert the changes related to the `buildName`. Though, to be clear, the resulting `pinia.browser-esm.prod.js` would be identical.

Another thing I'd note is that the minification of the file could be better, but it's easily good enough for now and I don't think that should be 'fixed' in this PR. I imagine this project will switch to tsdown at some point and (judging by Vue Router) that should give us a better set of minification defaults.

I have also taken a look at #2959. It's an interesting idea but I think it might need a bit more tweaking. In my opinion, this PR does what we need and should be merged first, rather than waiting for #2959.

posva:
Yes, migrating to tsdown would be great! I haven’t had the time to do it, so a PR is welcome if anybody has time

skirtles-code:
Work-in-progress migrating to tsdown: #3092.

PR #2947, opened 2025-03-18
feat(nuxt): move to esm-only

43081j:
Moves to an esm-only package now that all LTS other than 18.x (which is soon to fall out of LTS) support `require(esm)`.

This is a draft until nuxt/module-builder 1.x becomes non-alpha

1.x outputs esm only, whereas 0.x outputs cjs too.

### DIFF
--- a/packages/nuxt/package.json
+++ b/packages/nuxt/package.json
@@ -26,14 +26,10 @@
   "sideEffects": false,
   "type": "module",
   "exports": {
-    ".": {
-      "types": "./dist/types.d.ts",
-      "import": "./dist/module.mjs",
-      "require": "./dist/module.cjs"
-    }
+    ".": "./dist/module.mjs"
   },
-  "main": "./dist/module.cjs",
-  "types": "./dist/types.d.ts",
+  "main": "./dist/module.mjs",
+  "types": "./dist/module.d.mts",
   "files": [
     "dist"
   ],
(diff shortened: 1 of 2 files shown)


packages/nuxt/package.json
```diff
@@ -27,13 +27,11 @@
   "type": "module",
   "exports": {
     ".": {
-      "types": "./dist/types.d.ts",
-      "import": "./dist/module.mjs",
-      "require": "./dist/module.cjs"
+      "types": "./dist/types.d.mts",
+      "default": "./dist/module.mjs"
     }
```
posva: If I'm not mistaken you could even simplify this further!

https://github.com/posva/pinia-colada/blob/main/nuxt/package.json#L12

43081j: you're right. i assumed there wasn't a `module.d.mts` since this `types.d.mts` exists

but it seems like `types.d.mts` is just a wrapper around `module.d.mts` (i.e. it re-exports everything from it)

43081j:
according to the nuxt module-builder docs:

> types.d.ts: Exported types in addition to shims for nuxt.config auto completion.

but we don't reference that file anymore after this change.

maybe that's why the export had a separate `types`. in our case, though, `types.d.ts` just re-exports exactly what is in `module.mjs`

packages/nuxt/package.json
posva: I tested locally, and it seems to work with just the module.mjs, so let's add that!

43081j:
@danielroe do you know when nuxt-builder alpha is planned to become stable?

wondering if we should use the alpha tag for now or just wait

danielroe:
you should be safe to already use the alpha tag here - the only niggle is for modules that include Vue components in their `runtime/`.

posva (approved):

PR #2953, opened 2025-03-19
fix: devtools not showing up for nuxt 3.16

leon-marzahn:
Based on [this issue](https://github.com/nuxt/devtools/issues/823) in nuxt/devtools and [this discussion](https://github.com/vuejs/pinia/discussions/2952) here, after upgrading to nuxt 3.16 the devtools don't show up.

This is caused by the recent upgrade to their internal nuxt/devtools v2 which changes when the devtools are initialized. The registration only works correctly when the hook for it is added after vue devtools has setup up the subscribers, but with the new change in 3.16 the pinia plugin was loaded before that happens.

The main change is only registering the devtools after the app has mounted, which seems to be when vue devtools hooks are being subscribed based on hours of looking through their code.

https://github.com/vuejs/devtools/blob/d5f94279cf8e54fb05aa92209bcef767bb31b7c5/packages/devtools-kit/src/core/index.ts#L85C3-L107C5 
- This registers an app record that needs to be available before a devtool plugin is registered
- Incidentally it's also called after the actual store is registered, initially i had to check for a global property, but it turns out the property is already set by the time this hook is called
- The pinia nuxt plugin hook should be called after because it's registered later (nuxt app mounted, rather than app:init from vue)

I'm unsure of what needs to be tested or if we need to add anything or do anything because of the version change, but please let me know what i need to do. I'm also happy to downgrade the version here so we can support 3.15 but make it work in 3.16.

I've just had a long day trying to fix this 😅

### DIFF
--- a/packages/pinia/src/createPinia.ts
+++ b/packages/pinia/src/createPinia.ts
@@ -4,10 +4,14 @@ import { registerPiniaDevtools, devtoolsPlugin } from './devtools'
 import { IS_CLIENT } from './env'
 import { StateTree, StoreGeneric } from './types'
 
+export type CreatePiniaOptions = {
+  skipDevtoolsRegistration?: boolean
+}
+
 /**
  * Creates a Pinia instance to be used by the application
  */
-export function createPinia(): Pinia {
+export function createPinia(options?: CreatePiniaOptions): Pinia {
   const scope = effectScope(true)
   // NOTE: here we could check the window object for a state and directly set it
   // if there is anything like it with Vue 3 SSR
@@ -28,7 +32,7 @@ export function createPinia(): Pinia {
       app.provide(piniaSymbol, pinia)
       app.config.globalProperties.$pinia = pinia
       /* istanbul ignore else */
-      if (__USE_DEVTOOLS__ && IS_CLIENT) {
+      if (__USE_DEVTOOLS__ && IS_CLIENT && !options?.skipDevtoolsRegistration) {
         registerPiniaDevtools(app, pinia)
       }
       toBeInstalled.forEach((plugin) => _p.push(plugin))
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     ]
   },
   "resolutions": {
-    "@nuxt/kit": "^3.9.0",
-    "@nuxt/schema": "^3.9.0"
+    "@nuxt/kit": "^3.16.0",
+    "@nuxt/schema": "^3.16.0"
   }
 }
--- a/packages/nuxt/package.json
+++ b/packages/nuxt/package.json
@@ -46,7 +46,7 @@
     "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s --commit-path . -l @pinia/nuxt -r 1"
   },
   "dependencies": {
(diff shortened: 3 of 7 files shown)


packages/pinia/src/createPinia.ts
```diff
@@ -4,10 +4,14 @@ import { registerPiniaDevtools, devtoolsPlugin } from './devtools'
...
 
+export type CreatePiniaOptions = {
+  skipDevtoolsRegistration?: boolean
+}
+
 /**
  * Creates a Pinia instance to be used by the application
  */
-export function createPinia(): Pinia {
+export function createPinia(options?: CreatePiniaOptions): Pinia {
```
leon-marzahn, line 14: This will need to be documented but it shouldn't be a breaking change for anyone using it

posva:
Thanks for the PR. This is something that should be fixed in Nuxt devtools. In Vue devtools, they “collect” registrations that happen before the devtools are initialized and they set them up when possible.

leon-marzahn:
> Thanks for the PR. This is something that should be fixed in Nuxt devtools. In Vue devtools, they “collect” registrations that happen before the devtools are initialized and they set them up when possible.

The issue is that it does in fact still buffer registrations, but it only does so if the plugins get registered after the buffers have been initialized.

The issue with pinia's plugin is that it registers before they are initialized. There would be another way to fix it that would require the nuxt team to give the plugin a name so that we can load it afterwards, but that would also mean that it would load pinia after almost all other modules, which would break the pinia nuxt integration. 

Nuxt has a solution for this, which is the new devtools kit for specifically new nuxt devtools, but pinia/nuxt isn't using them.

I would say the 3 options are:
- Use a hack like this
- Rewrite the nuxt devtools to use their devkit and keep the Vue devtools integration separate
- Wait for the nuxt team to potentially provide a solution, but until then pinia doesn't support nuxt 3.16 which is a bit painful 

But i don't believe the nuxt team will fix this as they don't officially support vue devtools integrations that skip the nuxt devtools.

posva:
The devtools must work with both Vue and Nuxt. In this case, Nuxt must adapt to match Vue Devtools expectations. It will only be beneficial for the devtools to pick up native Vue devtools

leon-marzahn:
> The devtools must work with both Vue and Nuxt. In this case, Nuxt must adapt to match Vue Devtools expectations. It will only be beneficial for the devtools to pick up native Vue devtools

Nuxt only supports Vue 3. The setupDevtoolsPlugin is an API from Vue dev tools v6, which supports Vue 2. With the upgrade to Vue dev tools v7, they no longer support Vue 2. Nuxt now uses Vue dev tools v7, because they don't support Vue 2 and hence don't need to use Vue dev tools v6.

Pinia uses the Vue dev tools v6 API internally, why would Nuxt, which doesn't support the V6 api from Vue dev tools care about this?

The pinia/nuxt package needs to upgrade and stop using Vue dev tools v6, no other nuxt integration uses this API. Look at vueuse for example.

posva:
This got fixed in https://github.com/nuxt/devtools/commit/259853b94c38bf94e5ec7a6a18b426854becce75

PR #2954, opened 2025-03-23
feat: automatic HMR code (nuxt only)

BobbieGoede:
This is a quick and dirty implementation of a plugin that automatically appends the HMR code to files that register a pinia store (nuxt only), this only works for very basic setups in this state.

This could be implemented using unplugin so that it works in vue projects (without nuxt) as well, in a separate repo if preferred.

Maybe this functionality was already considered, I couldn't find an old discussion or PR suggesting this 🤔

### How it works

Given an example project has the following store:
```ts
// my-store.ts
export const myStore = defineStore('my-store', /* ... */)
```
The plugin checks if it contains `defineStore` and does not contain `acceptHMRUpdate`, then walks the top level nodes of the parsed code AST to find the export/variable declaration which uses the `defineStore` function. This code likely needs to be expanded to cover more use cases.

If an export/variable declaration is found, simply append the necessary HMR code:

```diff
// my-store.ts
export const myStore = defineStore('my-store', /* ... */)
+if (import.meta.hot) {
+  import.meta.hot.accept(
+    acceptHMRUpdate(myStore, import.meta.hot)
+  )
+}
```

## Summary by CodeRabbit

* **New Features**
  * Automatic Hot Module Replacement (HMR) support for stores during development.

* **Updates**
  * Counter now exposes a computed "double" value and the UI displays "Count: X x 2 = Y".
  * Main page shows raw counter state in a debug block for easier inspection.

### DIFF
--- a/packages/nuxt/src/auto-hmr-plugin.ts
+++ b/packages/nuxt/src/auto-hmr-plugin.ts
@@ -1,0 +1,66 @@
+import type { Nuxt } from 'nuxt/schema'
+
+function getStoreDeclaration(nodes?: import('estree').VariableDeclarator[]) {
+  return nodes?.find(
+    (x) =>
+      x.init?.type === 'CallExpression' &&
+      x.init.callee.type === 'Identifier' &&
+      x.init.callee.name === 'defineStore'
+  )
+}
+
+function nameFromDeclaration(node?: import('estree').VariableDeclarator) {
+  return node?.id.type === 'Identifier' && node.id.name
+}
+
+export function autoRegisterHMRPlugin(
+  nuxt: Nuxt,
+  { resolve }: { resolve: (...path: string[]) => string }
+) {
+  const projectBasePath = resolve(nuxt.options.rootDir)
+
+  return {
+    name: 'pinia:auto-hmr-registration',
+
+    transform(code, id) {
+      if (id.startsWith('\x00')) return
+      if (!id.startsWith(projectBasePath)) return
+      if (!code.includes('defineStore') || code.includes('acceptHMRUpdate')) {
+        return
+      }
+
+      const ast = this.parse(code)
+
+      // walk top-level nodes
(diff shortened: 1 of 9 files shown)


packages/nuxt/src/auto-hmr-plugin.ts
```diff
@@ -0,0 +1,64 @@
...
+                : undefined
+          )
+
+          // retrieve the variable name
+          const storeName = nameFromDeclaration(storeDeclaration)
+          if (storeName) {
+            // append HMR code
+            return {
+              code: [
+                `import { acceptHMRUpdate } from 'pinia'`,
```
posva, line 52: I added this so it works with `autoImports: false`. Let me know I overlooked anything

packages/nuxt/src/auto-hmr-plugin.ts
```diff
@@ -0,0 +1,64 @@
...
+                code,
+                'if (import.meta.hot) {',
+                `  import.meta.hot.accept(acceptHMRUpdate(${storeName}, import.meta.hot))`,
+                '}',
+              ].join('\n'),
+            }
```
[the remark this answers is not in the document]
posva: The current approach works

posva (approved):
Looks good to me. It doesn't seem to impact perf. I wonder if it will need any improvement to support rolldown later on. I had to add `filter` to `transform` in unplugin-vue-router for that

packages/nuxt/src/auto-hmr-plugin.ts
BobbieGoede: Sorry for the delay, these changes look good to me.

~~Might be worth checking if this works (or causes issues) with projects using webpack/rspack or if those needs some special treatment.~~

Nevermind, I just noticed the changelog says vite only and this is only added a vite plugin. I could look into making this work for those webpack/rspack if there's demand for it though.

PR #2959, opened 2025-03-31
build: add production build for every build type

### LINKED ISSUE #2205: Bundle ESM for Vue 3 prod

Doodidan:
Simplify build config generation to make production and minified version for every build type

Also make it to use same config logic for every build type so no more of hardcoding file name needed

Closes #2205 

Based on #2939

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -77,7 +77,7 @@ const packageConfigs = packageBuilds.map((format) =>
 packageBuilds.forEach((buildName) => {
   if (buildName === 'cjs') {
     packageConfigs.push(createProductionConfig(buildName))
-  } else if (buildName === 'global') {
+  } else if (buildName === 'global' || buildName === 'browser') {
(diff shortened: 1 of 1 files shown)


porfirioribeiro:
I was checking the build of this PR, I believe dev-tools dependency is being bundled, which is not ideal for prod usage

#2939 
```
-rw-r--r--@ 1 ribeipor  staff    69K May  2 12:12 pinia.esm-browser.js
-rw-r--r--@ 1 ribeipor  staff   6.2K May  2 12:12 pinia.esm-browser.prod.js
```

#2959
```
-rw-r--r--@ 1 ribeipor  staff    69K May  2 12:08 pinia.esm-browser.js
-rw-r--r--@ 1 ribeipor  staff    75K May  2 12:08 pinia.esm-browser.prod.js
```

Doodidan:
@porfirioribeiro Thank you, I have fixed this.

Doodidan:
#3092

PR #2960, opened 2025-04-01
feat(test): support unstubbing actions

CyberAP:
This change makes it possible to unstub actions when using `stubActions: true` in `createTestingPinia`. To do that you have to call `mockRestore` on the action.

```js
createTestingPinia()

useMyStore().someAction.mockRestore()

// this calls an actual implementation
useMyStore().someAction()
```

The support for unstubbing is provided only for jest and vitest.

### DIFF
--- a/packages/docs/cookbook/testing.md
+++ b/packages/docs/cookbook/testing.md
@@ -164,6 +164,33 @@ store.someAction()
 expect(store.someAction).toHaveBeenCalledTimes(1)
 ```
 
+If you want to normally execute only some of the actions with `stubActions: true` you can reset them with `mockRestore`:
+
+```js
+const wrapper = mount(Counter, {
+  global: {
+    // stubActions is set to 'true' by default
+    plugins: [createTestingPinia()],
+  },
+})
+
+const store = useSomeStore()
+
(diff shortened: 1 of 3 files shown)


lostncg:
This feature is useful, may I know when it gonna release? @posva

posva:
Thanks a lot for this! I'm reworking 
- https://github.com/vuejs/pinia/pull/3040
- to have https://github.com/vuejs/pinia/issues/2970 implemented

The approach you want should already be doable with type support using [the helper mentioned in docs](https://pinia.vuejs.org/cookbook/testing.html#Mocking-the-returned-value-of-an-action). It depends on your testing framework

CyberAP:
@posva this is somewhat different. We could stub specific actions before that:

```js
createTestingPinia({ stubActions: false })
actionsToStub.forEach(action => jest.spyOn(useMyStore(), action).mockReturnValue())
```

What we can't do is stub everything but unstub specific actions.

posva:
You can unstub specific actions, I just tested it. But now you can also pass a function to `stubActions` and return a boolean to decide wheather to stub or not. It should fit any advanced cases

PR #2965, opened 2025-04-04
docs: fix subscription arguments in docs

CyberAP:
Fixes a minor typo in subscription docs. The arguments should be `(mutation, state)`, not `(state)`.


posva (approved):
Thanks!

PR #2975, opened 2025-04-20
chore: split installation commands

MrMohamedAbdallah:
Currently, the installation commands for `yarn` and `npm` are combined in a single codeblock. When users click the copy button, both commands are copied together, forcing them to manually edit the pasted text. This PR splits them into separate codeblocks, allowing users to copy only the command they need, making the process more UX friendly.

### DIFF
--- a/packages/docs/getting-started.md
+++ b/packages/docs/getting-started.md
@@ -28,7 +28,6 @@ bun add pinia
 
 :::
 
-
 :::tip
 If your app is using Vue <2.7, you also need to install the composition api: `@vue/composition-api`. If you are using Nuxt, you should follow [these instructions](/ssr/nuxt.md).
 :::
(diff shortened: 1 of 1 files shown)


posva (changes requested):
Let’s make it like the one in Vue router docs

packages/docs/getting-started.md
```diff
@@ -7,12 +7,20 @@
...
+```bash [npm]
 npm install pinia
 ```
 
+```bash [yarn]
+yarn add pinia
+```
+
+:::
+
```
posva, line 24:
```suggestion
```

posva (changes requested):
Let's add the pnpm and bun commands too (like in vue router docs)

posva (approved):
Thanks!

PR #2976, opened 2025-04-20
chore: split installation commands in README.md

MrMohamedAbdallah:
Make it easier to copy and paste installation commands with GitHub "copy" button.


posva:
Thanks but i prefer to keep it simple in the readme

PR #2978, opened 2025-04-25
chore: Support Nuxt 4

amonys19212:

### DIFF
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -27,7 +27,7 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
     name: 'pinia',
     configKey: 'pinia',
     compatibility: {
-      nuxt: '>=3.15.0',
+      nuxt: '^3.15.0 || ^4.0.0-0',
     },
   },
   defaults: {},
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -27,7 +27,7 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
     name: 'pinia',
     configKey: 'pinia',
     compatibility: {
-      nuxt: '^3.15.0',
+      nuxt: '>=3.15.0',
     },
   },
   defaults: {},


posva:
Thanks but only once Nuxt 4 is released, it should be tested and the version adapted:

CleanShot 2025-04-25 at 14 11 44

amonys19212:
Thank you very much for your reply. I am currently testing Nuxt 4 (Nightly Channel). Unfortunately, due to the module’s version settings, nuxt/pinia is not able to run on Nuxt 4 at the moment. However, the older version (v0.9.0) is able to run successfully.


**Example**  Nuxt v4.0.0-29092227.317ced76 + pinia v2.3.1 + @pinia/nuxt v0.9.0
https://stackblitz.com/edit/nuxt-starter-zayjogfc


https://www.npmjs.com/package/nuxt-nightly?activeTab=versions

packages/nuxt/src/module.ts
```diff
@@ -27,7 +27,7 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
     name: 'pinia',
     configKey: 'pinia',
     compatibility: {
-      nuxt: '^3.15.0',
+      nuxt: '>=3.15.0',
```
posva: Let's keep it a bit stricter:

```suggestion
      nuxt: '^3.15.0 || ^4.0.0-0',
```

posva (commented):
I wasn't aware of the nuxt-nightly having a v4 available!
Thanks for testing it out!

posva:
You should be able to use the _publish any commit_ version and polish the PR as needed

packages/nuxt/src/module.ts
amonys19212: You’re right, thank you very much!

nathanchase:
This should get probably get priority now that 4.0 is in alpha: https://github.com/nuxt/nuxt/releases/tag/v4.0.0-alpha.1

```
 WARN  Module pinia is disabled due to incompatibility issues:                                            
 - [nuxt] Nuxt version ^3.15.0 is required but currently using 4.0.0-alpha.1
```

posva (approved):
I tested locally and it works. It does warn

```
 WARN  [nuxt] Expected @nuxt/kit to be at least 4.0.0-alpha.1 but got 3.17.5. This might lead to unexpected behavior. Check your package.
json or refresh your lockfile.
```

But I suppose this is expected in the meantime

PR #2979, opened 2025-04-28
fix(docs): Props destructuring suggestion is not true anymore with the release of vue 3.5

michaelcozzolino:
Constrained the hint to vue < 3.5, because from 3.5 props destructuring is reactive

### DIFF
--- a/packages/docs/core-concepts/index.md
+++ b/packages/docs/core-concepts/index.md
@@ -136,7 +136,7 @@ import { computed } from 'vue'
 
 const store = useCounterStore()
 // ❌ This won't work because it breaks reactivity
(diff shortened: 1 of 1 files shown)


packages/docs/core-concepts/index.md
```diff
@@ -136,7 +136,7 @@ import { computed } from 'vue'
 
 const store = useCounterStore()
 // ❌ This won't work because it breaks reactivity
-// it's the same as destructuring from `props`
+// it's the same as destructuring from `props` for Vue < 3.5
```
```suggestion
// same as reactive: https://vuejs.org/guide/essentials/reactivity-fundamentals.html#limitations-of-reactive
```

PR #2983, opened 2025-05-27
feat(warn): detect global context on the server side

ivansky:
when store is accessed and injection context is not found it falls back to global active pinia object, 

although it is safe for client, on the server side there are could be many requests with their own context,

falling back to global state leads to race conditions and undefined behaviour,

this should be fixed once and for all!

it is probably a breaking change but a good one, I haven't tested this locally, I appreciate any help to improve this PR in order to make it good looking

## Summary by CodeRabbit

* **New Features**
  * Added public APIs to get and set the active Pinia instance.

* **Tests**
  * Added SSR tests validating active-Pinia behavior when called outside a Pinia context.
  * Enhanced console mock utilities to track and assert on both warnings and errors (new error-focused helpers and matchers).

* **Improvements**
  * Development-mode diagnostics now warn about potential SSR / cross-request Pinia context issues.

### DIFF
--- a/packages/pinia/src/rootStore.ts
+++ b/packages/pinia/src/rootStore.ts
@@ -6,7 +6,6 @@ import {
   InjectionKey,
   Ref,
 } from 'vue'
-import { createPinia } from './createPinia'
 import {
   StateTree,
   PiniaCustomProperties,
@@ -18,6 +17,7 @@ import {
   DefineStoreOptionsInPlugin,
   StoreGeneric,
 } from './types'
+import { IS_CLIENT } from './env'
 
 /**
  * setActivePinia must be called to handle SSR at the top of functions like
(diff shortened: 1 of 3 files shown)


packages/pinia/src/rootStore.ts
```diff
@@ -43,7 +43,7 @@ interface _SetActivePinia {
  * Get the currently active pinia if there is any.
  */
 export const getActivePinia = () =>
-  (hasInjectionContext() && inject(piniaSymbol)) || activePinia
+  (hasInjectionContext() && inject(piniaSymbol)) || (import.meta.server ? throw new Error("Cannot get active pinia as it does not find context") : activePinia)
```
posva: Thanks for the PR. This could be changed into a non breaking change by just showing an error and still returning the active pinia. The error should be in development only (like other warnings) and should not rely on `import` as it might not work in all scenarios

posva (changes requested):

packages/pinia/src/rootStore.ts
ivansky: I think failing fast is the better approach when the global `activePinia` is accessed on the server, because:
1. failing immediately makes the problem obvious rather than silently continuing with potentially incorrect behavior.
2. allowing it to continue by returning `activePinia` could lead to dangerous state sharing between requests, which is particularly problematic in server environments.

maybe making the message is more clear would help huge

posva: That would be a breaking change. Also, I do prefer the error message because it feels less frustrating to users

ivansky: I will leave the decision to you,

However, please keep in mind that the developers tend to ignore error messages, so they will make the same mistakes again and again, I would prefer it to fail that will indicate the mistake much faster.

ivansky:
@posva I removed throwing error, added `console.error` and fallback to `createPinia()` instead, please check.

iPrytz:
Perfect @ivansky ! This would be great to have at least a warning. Would also have preferred the throw error but anything is better than nothing.

posva (approved):

PR #3004, opened 2025-07-15
fix: can't access property "size", store._customProperties is undefined

TimvdEijnden:
When using Pinia module in the Vue dev tools in Firefox (Zen browser) and I select a store the following error occurs, it is fixed by checking if `store._customProperties` exists.

Screenshot 2025-07-15 at 13 48 36

When I set a breakpoint and set `store._customProperties` to a new Map() it seems to work.

Screenshot 2025-07-15 at 13 48 21

### DIFF
--- a/packages/pinia/src/devtools/formatting.ts
+++ b/packages/pinia/src/devtools/formatting.ts
@@ -159,7 +159,7 @@ export function formatStoreForInspectorState(
     }))
   }
 
-  if (store._customProperties.size) {
+  if (store._customProperties && store._customProperties.size) {
     state.customProperties = Array.from(store._customProperties).map((key) => ({
       editable: true,
       key,


posva:
do you have a repro?

TimvdEijnden:
I can reproduce it on https://www.jumbo.com but I can try create a separate one if really needed

posva:
Yes, please, a repro must be minimal

TimvdEijnden:
Here it is the repro: https://stackblitz.com/edit/nuxt-pinia-customproperties-undefined?file=app.vue

posva:
thanks! That's quite weird, I suppose it goes away if you use the nuxt devtools?
Regarding the PR, this shouldn't be the fix as the _customProperties should always be defined, so if you want to push this forward, you will need to dig a bit deeper into the problem. Thanks!

TimvdEijnden:
[Here](https://github.com/vuejs/pinia/blob/31a0c97b0246dcb3d047ab61c3b71cc68dbca6ab/packages/pinia/src/store.ts#L476) you see the `_customProperties` only added in case ` __DEV__ || (__USE_DEVTOOLS__ && IS_CLIENT)` and that is false in this scenario. 

```ts
const store: Store<Id, S, G, A> = reactive(
    __DEV__ || (__USE_DEVTOOLS__ && IS_CLIENT)
      ? assign(
          {
            _hmrPayload,
            _customProperties: markRaw(new Set<string>()), // devtools custom properties
          },
          partialStore
          // must be added later
          // setupStore
        )
      : partialStore
  ) as unknown as Store<Id, S, G, A>
```

I only wanted to get rid of the error, not sure what the best best appoach is to fix it since the app is running production mode.

PR #3009, opened 2025-07-17
fix(nuxt): mark compatibility for nuxt v4 stable

### LINKED ISSUE #3008: Pinia 3.0.3 is not compitable with Nuxt 4.0
### Reproduction

Upgrade to Nuxt 4.0

Hi!

For some reason Pinia is no longer compitable with Nuxt 4.0 that while there were no issues with its Alpha counter part.

 WARN  Module pinia is disabled due to incompatibility issues:                                                                                              
 - [nuxt] Nuxt version ^3.15.0 is required but currently using 4.0.0

### Expected behavior

The module should work without warning or getting disabled.

### Actual behavior

The module gets disabled

### Additional information

pinia: 3.0.3
nuxt: 4.0.0

GalacticHypernova:
Resolves #3008 

Nuxt v4 (the non-alpha version) has a version that the current compatibility string doesn't capture. This adds the new stable version to the compatibility string (we may want to combine the 2 v4 strings)

### DIFF
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -27,7 +27,7 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
     name: 'pinia',
     configKey: 'pinia',
     compatibility: {
-      nuxt: '^3.15.0 || ^4.0.0-0 || ^4.0.0',
+      nuxt: '^3.15.0 || ^4.0.0',
     },
   },
   defaults: {},
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -27,7 +27,7 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
     name: 'pinia',
     configKey: 'pinia',
     compatibility: {
-      nuxt: '^3.15.0 || ^4.0.0-0',
+      nuxt: '^3.15.0 || ^4.0.0-0 || ^4.0.0',
     },
   },
   defaults: {},


packages/nuxt/src/module.ts
```diff
@@ -27,7 +27,7 @@ const module: NuxtModule<ModuleOptions> = defineNuxtModule<ModuleOptions>({
     name: 'pinia',
     configKey: 'pinia',
     compatibility: {
-      nuxt: '^3.15.0 || ^4.0.0-0',
+      nuxt: '^3.15.0 || ^4.0.0-0 || ^4.0.0',
```
posva: Thanks! This might be a bug in the semver check instead: https://jubianchi.github.io/semver-check/#/^4.0.0-0/4.0.0

GalacticHypernova: It seems like we can simplify it further even
https://jubianchi.github.io/semver-check/#/^4.0.0/4.0.0-alpha

posva: Interesting! That didn't use to be the case with npm 👀

GalacticHypernova: Indeed it is! But at least its a simpler syntax, so no harm in adopting it =)

Barbapapazes: This is really strange that it works because 4.0.0-alpha is clearly behind a 4.0.0 release and `^` means 'after' on the second and third digit.

Barbapapazes: image

`4.0.0-alpha` isn't in this range.

GalacticHypernova: @Barbapapazes in that case, could it actually be a display error on semver's end? I guess we should support stable releases more than alpha pre-releases, but maybe in order to support both we would need to combine the previous compatibility string with this PR's?

posva: The current version with just the stable is fine

PR #3037, opened 2025-09-07
feat: judge if action return value is promise like

Hermes-Axe:
## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures actions that return promise-like values (thenables) are correctly recognized and handled.
  * Fixes error propagation for asynchronous actions, routing failures to error handlers and returning rejected promises as expected.
  * Preserves success behavior by triggering completion callbacks with resolved values for both Promises and thenables.
* **Reliability Improvements**
  * Provides more consistent async behavior across different promise-like returns, reducing edge-case failures in action handling.

### DIFF
--- a/packages/pinia/src/store.ts
+++ b/packages/pinia/src/store.ts
@@ -75,6 +75,10 @@ interface MarkedAction<Fn extends _Method = _Method> {
   [ACTION_NAME]: string
 }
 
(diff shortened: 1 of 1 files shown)


posva:
Thank you but this code is intended. You should use native promises not custom wrappers

PR #3041, opened 2025-09-14
fix: trigger reactivity for shallowRef when using  object syntax

### LINKED ISSUE #2861: Mutating a shallowRef object using $patch doesn't trigger reactivity
### Reproduction

https://stackblitz.com/edit/github-5ayhfs4m?file=src%2FApp.vue

1. Define a shallowRef state (e.g. `const counter = shallowRef({ count: 0 })`) using setup stores syntax
2. Use the store in a Vue component
3. Update the state using `$patch({ counter: { count: 1 })` syntax
4. The state is updated in store but the reactivity is not triggered

### Expected behavior

At step 4, the Vue's reactivity system should trigger

### Actual behavior

At step 4, the Vue's reactivity system does not trigger (no watcher/computed run after the state change)

### Additional information

In the playground, I tested with two other ways to update the shallowRef's state, and the reactivity still works normally:
```ts
const counterStore = useCounterStore();
// Using $patch with update function: trigger reactivity
counterStore.$patch((state) => {
    state.counter2 = { count: state.counter2.count + 1 };
  });
// Direct mutate the state: trigger reactivity
counterStore.counter3 = { count: ... };
```

doubledare704:
- Add triggerRef import from Vue
- Implement isShallowRef utility function to detect shallowRef using __v_isShallow
- Modify  method to detect shallowRef targets and trigger reactivity after patching
- Add comprehensive tests for shallowRef reactivity scenarios
- Ensure compatibility with existing patch behavior

close #2861

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored reliable reactivity for shallowRef state when using $patch — nested, partial, multi-field, and direct-assignment updates now consistently notify watchers and computed values; public APIs unchanged.

* **Tests**
  * Added comprehensive tests covering shallowRef behavior with object and function patch syntax, nested updates, partial/multi-ref patches, direct assignments, and reactivity observation to prevent regressions.

### DIFF
--- a/packages/pinia/__tests__/store.patch.spec.ts
+++ b/packages/pinia/__tests__/store.patch.spec.ts
@@ -224,8 +224,6 @@ describe('store.$patch', () => {
       setActivePinia(createPinia())
       return defineStore('shallowRef', () => {
         const counter = shallowRef({ count: 0 })
-        const counter2 = shallowRef({ count: 0 })
-        const counter3 = shallowRef({ count: 0 })
         const markedRaw = ref({
           marked: markRaw({ count: 0 }),
         })
@@ -237,8 +235,6 @@ describe('store.$patch', () => {
         return {
           markedRaw,
           counter,
-          counter2,
-          counter3,
           nestedCounter,
         }
       })()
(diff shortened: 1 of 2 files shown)


doubledare704:
@coderabbitai generate docstrings

PR #3042, opened 2025-09-14
feat: expose store options in PiniaCustomProperties for plugin access

### LINKED ISSUE #1247: Plugins: Expose `DefineStoreOptionsBase` in `PiniaCustomProperties`

I wrote a plugin to solve #625. This is basically working, but the types of the stores defined in the new stores option in `DefineStoreOptionsBase` are not accessible inside `PiniaCustomProperties`.

```ts
import type {PiniaPluginContext, Store, StoreDefinition} from 'pinia';

declare module 'pinia' {
  export interface DefineStoreOptionsBase<S, Store> {
    stores?: Record<string, StoreDefinition>;
  }

  export interface PiniaCustomProperties<Id, S, G, A> {
    readonly stores: Readonly<Record<string, Store>>;
  }
}

export function piniaPluginStores (context: PiniaPluginContext) {
  return {
    get stores () {
      return Object.freeze(
        Object.entries(context.options.stores ?? {})
          .reduce<Record<string, Store>>((accumulator, [name, definition]) => {
            accumulator[name] = definition();
            return accumulator;
          }, {}),
      );
    },
  };
}
```

### Proposed solution

If the store options would be accessible inside `PiniaCustomProperties`, the actual types of the concrete stores could be accessed with `this.stores`.

_No response_

doubledare704:
- Add StoreOptionsAccess utility type for accessing custom store options
- Modify store creation to include _options property with store options
- Export StoreOptionsAccess type from main module
- Add comprehensive tests for plugin access to custom store options
- Support both option stores and setup stores
- Maintain backward compatibility with existing plugins

Fixes #1247

## Summary by CodeRabbit

* **New Features**
  * Store instances now expose per-store custom options to plugins (works for both options- and setup-style stores) while preserving backward compatibility.
  * Added a public TypeScript utility type to surface store option typings for plugins and consumers.

* **Tests**
  * Added comprehensive runtime and type-level tests validating plugin access, typings, and behavior across multiple store configurations.

### DIFF
--- a/packages/pinia/__tests__/storeOptionsAccess.spec.ts
+++ b/packages/pinia/__tests__/storeOptionsAccess.spec.ts
@@ -1,0 +1,264 @@
+import { describe, it, expect } from 'vitest'
+import { createPinia, defineStore, StoreDefinition } from '../src'
+import { mount } from '@vue/test-utils'
+import { ref } from 'vue'
+
+// Extend the types to test the new functionality
+declare module '../src' {
+  export interface DefineStoreOptionsBase<S, Store> {
+    stores?: Record<string, StoreDefinition>
+    customOption?: string
+    debounce?: Record<string, number>
+  }
+
+  export interface PiniaCustomProperties<Id, S, G, A> {
+    readonly stores: any
+    readonly customOption: any
+    readonly debounce: any
+  }
+}
+
+describe('Store Options Access', () => {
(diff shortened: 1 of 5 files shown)


doubledare704:
@coderabbitai generate docstrings

posva:
Thanks for the PR. While it does look like fully AI generated and I don't think it solves the actual issue correctly, I appreciate the effort.

PR #3052, opened 2025-10-02
Update composing-stores.md

Jiamingou:
According to "https://pinia.vuejs.org/core-concepts/#Setup-Stores", it claimed that  "you must return all state properties in setup stores for Pinia to pick them up as state." Anyway, this code broke that rule , I believe is a typo, hence I "put" the `list` ref back into the user store.

## Summary by CodeRabbit

- Documentation
  - Updated the composing stores guide to show the cart example returning a list alongside summary and purchase.
  - Examples now reference a single source of truth for list data, improving clarity and consistency.
  - Clarified how list length and data are derived to better match real usage.

### DIFF
--- a/packages/docs/cookbook/composing-stores.md
+++ b/packages/docs/cookbook/composing-stores.md
@@ -55,16 +55,17 @@ import { apiPurchase } from './api'
 
 export const useCartStore = defineStore('cart', () => {
   const user = useUserStore()
+  const list = ref([])
 
   const summary = computed(() => {
-    return `Hi ${user.name}, you have ${user.list.length} items in your cart. It costs ${price.value}.`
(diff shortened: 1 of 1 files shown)


packages/docs/cookbook/composing-stores.md
```diff
@@ -55,14 +55,13 @@ import { apiPurchase } from './api'
 
 export const useCartStore = defineStore('cart', () => {
   const user = useUserStore()
-  const list = ref([])
```
posva, line 58: It shouldn't be removed, it should just be returned at the end

posva (changes requested):

posva (approved):

Issue #2914, opened 2025-02-15
3.0.1,3.0.2,3.0.3 - devtoolsApi is not defined


lianee:
### Reproduction

https://jsfiddle.net/0sj7apb6/

### Steps to reproduce the bug

Using the pinia.iife.js 3.0.1 now gives the error: "Uncaught ReferenceError: devtoolsApi is not defined"
no error with 3.0.0, probably related to #2910 fix

### Expected behavior

no error, and working devtools with the no prod version

### Actual behavior

error using the iife.js, and no devtools using the iife.prod.js

posva:
The devtools API has indeed been removed from the iifee given the considerable size increase. It has to be included manually. This should be added to docs (migration guide) and to the changelog.

lianee:
I understand, but I haven't found how to include it manually in iife. I just see a cjs and a esm version in node_modules/@vue/devtools-api/lib/
Perhaps this lib should be included in the dev version (iife.js), and not in iife.prod.js, I suspect it was like that in 3.0.0, and I don't see a problem having a big dev file

houd1ni:
I'm having the similar issue, isn't it related? I'm migrating from 2.3.1 to 3.0.1. And I'm not using any devtools related APIs at all, also even disabled devtools in my browser to check with no success.

```
Uncaught TypeError: Cannot create proxy with a non-object as target or handler
    at node_modules/@vue/devtools-kit/dist/index.js (index_f240.0be6a908.js:2088:34)
    at ModuleSystem.require (07032025/:395:26)
    at node_modules/pinia/node_modules/@vue/devtools-api/dist/index.js (index_f290.b725f0c7.js:2054:27)
    at ModuleSystem.require (07032025/:395:26)
    at node_modules/pinia/dist/pinia.mjs (index_acf3.09fbed4d.js:927:27)
    at ModuleSystem.require (07032025/:395:26)
    at src/store.ts (router.tsx:26:1)
    at ModuleSystem.require (07032025/:395:26)
    at src/styles.ts (store.ts:21:1)
    at ModuleSystem.require (07032025/:395:26)
```

lianee:
Unfortunately, this problem persists with version 3.0.2, rendering the non-prod version unusable with iife build.

I'm afraid I wasn't clear enough: these versions of pinia are no longer usable in development if you integrate pinia directly into the browser (without a package manager)

[...]

PR #3072, opened 2025-11-23
Remove broken IIFE build 

tux-tn:
Pinia’s IIFE build is unusable because `@vue/devtools-api` doesn’t ship an IIFE (see #2914). This PR drops the broken output from distributed formats and points unpkg/jsfiddle to the esm-browser build instead.

## Summary by CodeRabbit

* **Chores**
  * Updated CDN distribution formats to use modern ESM standard for unpkg and jsdelivr deliveries.
  * Removed legacy IIFE build configuration.

### DIFF
--- a/packages/pinia/package.json
+++ b/packages/pinia/package.json
@@ -5,8 +5,8 @@
   "type": "module",
   "main": "index.cjs",
   "module": "dist/pinia.mjs",
-  "unpkg": "dist/pinia.iife.js",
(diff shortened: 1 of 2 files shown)


posva:
Thanks, I might remove the iife build in the future in favor of the esm version and import maps but for the moment this was fixed in devtools repo (see https://github.com/vuejs/pinia/issues/2914#issuecomment-4959306305)

PR #3075, opened 2025-11-24
feat(dx): warn when storeToRefs is called with a plain object

### LINKED ISSUE #3074: feat(dx): warn when storeToRefs() receives a plain object
### Reproduction

This is not a runtime bug, but a DX issue.  ```ts import { defineStore, storeToRefs } from 'pinia'  const useCounter = defineStore('counter', { state: () => ({ count: 0 }) }) const counter = useCounter()  // ❌ Wrong usage (common beginner mistake) const wrong = { count: counter.count }  const { count } = storeToRefs(wrong) // `count` is NOT reactive — silent failure

Create any Pinia store.

Pass a plain object to storeToRefs(), for example:
storeToRefs({ count: store.count }).

Notice that no warning is shown.

The returned refs are not reactive — reactivity silently breaks.

### Expected behavior

In dev mode, storeToRefs() should show a small warning when the argument
is clearly not a Pinia store (missing $id and not reactive).

This helps beginners quickly identify a common misuse

### Actual behavior

storeToRefs() accepts plain objects without warning.

This causes confusing bugs because the returned refs do not update
and appear "stuck", with no indication of the mistake.

### Additional information

Proposal:

Add a small dev-only check inside storeToRefs():

If the input has no $id

And is not a ref

And is not reactive

Then show a simple warning message in DEV mode.

This is 100% backward compatible and has no impact on production.

VIPAX-JIT:
### Linked Issue
Fixes: #3074

### Description

This PR adds a development-only warning when `storeToRefs()` is called with a plain object instead of a Pinia store or a reactive object. This is a common beginner mistake and currently fails silently, producing refs that are not reactive. Adding a warning improves DX by immediately pointing out the misuse.

### What it handles

Example of incorrect usage:

```ts
const useCounter = defineStore('counter', { state: () => ({ count: 0 }) })
const counter = useCounter()

// ❌ not a store instance
const { count } = storeToRefs({ count: counter.count })
```

This returns a ref that is not connected to the actual store state.

This PR adds a minimal `__DEV__` check to detect this case and show a warning:

- If the input has no `$id`
- And is not a ref
- And is not reactive

### Tests

Added a test in `packages/pinia/__tests__/storeToRefs.spec.ts`:

```ts
it('warns if a plain object is passed', () => {
  const spy = vi.spyOn(console, 'warn').mockImplementation(() => {})
  storeToRefs({} as any)
  expect(spy).toHaveBeenCalledTimes(1)
  spy.mockRestore()
})
```

### Safety

- This is **purely dev-only** (`__DEV__` block)
- No breaking changes
- No behavioral changes in production
- Fully backward-compatible

Happy to update anything based on feedback!

## Summary by CodeRabbit

* **Tests**
  * Added test case validating storeToRefs warning behavior.

* **New Features**
  * storeToRefs now displays a console warning when called with invalid input.

### DIFF
--- a/packages/pinia/__tests__/storeToRefs.spec.ts
+++ b/packages/pinia/__tests__/storeToRefs.spec.ts
@@ -200,6 +200,16 @@ describe('storeToRefs', () => {
     expect(spy).toHaveBeenCalledTimes(0)
   })
 
+  it('warns if a plain object is passed', () => {
+    const spy = vi.spyOn(console, 'warn').mockImplementation(() => {})
+    storeToRefs({} as any)
+    expect(spy).toHaveBeenCalledTimes(1)
+    expect(spy).toHaveBeenCalledWith(
+      expect.stringContaining('storeToRefs() expects a Pinia store')
+    )
+    spy.mockRestore()
+  })
+
   tds(() => {
     const store1 = defineStore('a', () => {
       const n = ref(0)
--- a/packages/pinia/src/storeToRefs.ts
+++ b/packages/pinia/src/storeToRefs.ts
@@ -87,6 +87,12 @@ export type StoreToRefs<SS extends StoreGeneric> =
 export function storeToRefs<SS extends StoreGeneric>(
   store: SS
 ): StoreToRefs<SS> {
+  if (__DEV__ && !isRef(store) && !isReactive(store) && !('$id' in store)) {
+    console.warn(
+      `[🍍]: storeToRefs() expects a Pinia store but got a plain object.`
+    )
+  }
+
   const rawStore = toRaw(store)
 
   const refs = {} as StoreToRefs<SS>


packages/pinia/__tests__/storeToRefs.spec.ts
```diff
@@ -200,6 +200,16 @@ describe('storeToRefs', () => {
     expect(spy).toHaveBeenCalledTimes(0)
   })
 
+  it('warns if a plain object is passed', () => {
+    const spy = vi.spyOn(console, 'warn').mockImplementation(() => {})
+    storeToRefs({} as any)
```
posva, line 205: This is why this is not needed

posva:
Thanks for the PR but I don't think this is needed

packages/pinia/__tests__/storeToRefs.spec.ts
VIPAX-JIT: Hi, thanks for the clarification. I understand that adding a warning for passing a plain object to storeToRefs is unnecessary since this is already prevented by proper usage and TypeScript typings.

I’ve removed the warning logic and the related test from the PR, keeping the implementation aligned with Pinia’s intended behavior. Please let me know if further adjustments are needed.
Thanks for the guidance!

PR #3092, opened 2026-01-10
build: use tsdown

### LINKED ISSUE #2205: Bundle ESM for Vue 3 prod

skirtles-code:
Close #2939 

Migrate the main build to tsdown.

Marking as a draft for now. I believe everything is working correctly apart from type generation.

---

**Old problem (now fixed):**

`test:dts` is failing. It seems the problem is related to generics being renamed. For example, we're getting this in the generated `.d.ts`:

```ts
interface PiniaCustomProperties<Id$1 extends string = string, S$1 extends StateTree = StateTree, G$1 = _GettersTree<S$1>, A = _ActionsTree> {}
```

The original source code calls those generics `Id`, `S`, `G`, without the `$1` suffixes. Removing those suffixes fixes the TS errors.

I believe this might be the underlying issue:

- https://github.com/rolldown/rolldown/issues/6586

This was fixed in `1.0.0-beta.60`.

---

**Current problem:**

We have a single test failure. `customizations.test-d.ts` is accessing `StoreActions` without importing it, which worked previously but doesn't work now. I've confirmed that this fails in a 'real' application, it isn't a problem with the test itself.

I asked about this on the TypeScript Discord:

-  https://discord.com/channels/508357248330760243/1466086808231350531
 
The suggestion there is that this *could* be a TS bug. So I opened an issue here:

- https://github.com/microsoft/TypeScript/issues/63061

I haven't yet received confirmation about whether this is a bug, but it doesn't really matter, as it's still going to be a problem for Pinia either way. I've raised a discussion on the Rolldown Discord, but haven't yet received any feedback:

- https://discord.com/channels/1171783886200643635/1215661161329065987/1466466230226391133

I'll wait a bit longer before trying to take the further.

## Summary by CodeRabbit

* **Chores**
  * Removed TypeScript declaration build step from CI pipeline
  * Cleaned up unused imports and global type declarations
  * Removed unused configuration files

* **Build & Tooling**
  * Modernized build system for improved performance
  * Enhanced package metadata (keywords, license, repository information)


posva:
THe renaming is interesting. I think generic params must keep the same name for some module augmentation to work.

skirtles-code:
The renaming problem is now fixed. `1.0.0-beta.60` was an incomplete fix, but it did work well enough for Pinia.

There's still a failure in the dts tests. It seems that `customizations.test-d.ts` is accessing `StoreActions` without importing it, which worked before but fails now. I'm not clear exactly how this aspect of TS works, so I'll need to do a bit of digging to try to understand whether this is our problem or a tsdown problem.

posva (approved):
This saved me so much time! ❤️

PR #3095, opened 2026-01-17
build: fix changelog

skirtles-code:
I've tried to fix some problems in the `CHANGELOG.md` for the main Pinia package:

1. Several versions from 2.2.7 onwards were listed twice, though sometimes with small differences between the duplicates. I've removed the extra entries and tried to merge any important parts from the removed entries back into the retained entries.
2. The links in the headings were missing or broken for versions since 2.2.7. The tag used for releases was changed in that version from `pinia@X.Y.Z` to `vX.Y.Z` and the headings were still using the old convention.
3. Recent headings have all been level 3 (i.e. `###`). This doesn't make sense as the versions should have a lower heading depth than the sub-headings such as `Features` and `Bug Fixes`. I've changed some of the headings to `#` or `##` to reflect the conventions used for earlier builds (level 1 for majors and minors, level 2 for patches). I'm unclear why `conventional-changelog` suddenly started generating level 3 headings for everything.

The first two problems listed above appear to be because `conventional-changelog` was using the wrong tag prefix for Pinia. It was looking for tags beginning with `pinia@`, which led to it finding 2.2.6, rather than the latest release. This seems to be caused by the use of the `--lerna-package` flag, which sets the tag prefix. See:

- https://github.com/conventional-changelog/conventional-changelog/blob/0c786fdf1e4ba289a60d21619cdca88299e36c0a/packages/conventional-changelog/src/cli/options.ts#L41

I'm not very familiar with that package, or how it is being used by `release.mjs`, so it's very possible that my proposed fix is not correct. In my testing it did seem to fix the first two problems mentioned earlier.

## Summary by CodeRabbit

* **Documentation**
  * Reformatted changelog with updated heading levels and link formats across version history; reorganized and refined breaking changes notes.

* **Chores**
  * Updated release script to optimize changelog generation for package builds.

### DIFF
--- a/packages/pinia/CHANGELOG.md
+++ b/packages/pinia/CHANGELOG.md
@@ -1,4 +1,4 @@
-### [3.0.4](https://github.com/vuejs/pinia/compare/pinia@2.2.6...pinia@3.0.4) (2025-11-05)
+## [3.0.4](https://github.com/vuejs/pinia/compare/v3.0.3...v3.0.4) (2025-11-05)
 
 ### Features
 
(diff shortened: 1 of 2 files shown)


posva (commented):
Thanks a lot! I need to replace the release script with a new one that uses trusted publishing so I will merge this the next time I need to release a new version

posva (approved):
Thanks, this is so helpful!

PR #3098, opened 2026-01-26
refactor(types): remove Vue 2 type extensions

skirtles-code:
I believe these types are only relevant to Vue 2, so they can be removed from Pinia 3.

There is still a `@ts-ignore` on the `declare module 'vue'` (lines 13 & 14), which I haven't changed. It might be possible to remove that `@ts-ignore`, but I wasn't confident that'd be safe to change so I've left it in.

## Summary by CodeRabbit

* **Refactor**
  * Removed Vue 2-specific type augmentations
  * Simplified type imports and internal type definitions
  * Vue 3 API remains unchanged


posva (approved):
Thanks! This is the danger of ts-ignore

PR #3082, opened 2025-12-05
refactor: use Object.prototype.hasOwnProperty.call() for safety

curiouscoder-cmd:
## Summary

This PR replaces direct `obj.hasOwnProperty(key)` calls with the safer `Object.prototype.hasOwnProperty.call(obj, key)` pattern.
fixes #3083 
### Why?

Calling `hasOwnProperty` directly on an object can fail if:
- The object was created with `Object.create(null)` (no prototype)
- The object has a property named `hasOwnProperty` that shadows the method

The codebase already uses the safe pattern in `shouldHydrate()` function. This PR applies the same pattern consistently to the `mergeReactiveObjects()` function in both:

### Changes
- **packages/pinia/src/store.ts**: 2 occurrences fixed
- **packages/testing/src/testing.ts**: 2 occurrences fixed

### Type of change
- [x] Refactor (no functional changes)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved property validation robustness in core modules to handle edge cases with custom object prototypes, ensuring more reliable behavior with complex object structures.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #3101, opened 2026-02-09
fix: use safe hasOwnProperty call when merging patches

PliskoBogdan:
### What changed?

Replaced direct `patchToApply.hasOwnProperty()` usage with
`Object.prototype.hasOwnProperty.call()` inside `mergeReactiveObjects`.

### Why?

Calling `hasOwnProperty` directly can be unsafe if the object overrides the method.
Using the prototype call is the recommended safe approach.

## Summary by CodeRabbit

* **Refactor**
  * Improved code reliability through enhanced property access checks in the store module.


posva:
Thanks, but duplicate of #3082

PR #3102, opened 2026-02-10
Docs: fix invalid mapState example and magicValue usage

tandddian:
# This PR fixes two issues in the `mapState` example in `state.md`:

1. Add a missing trailing comma after `...mapState(...)` to avoid a syntax error when copying the example.
2. Fix `magicValue` to use `store.count` instead of `this.count`, which is not reliably defined on the component instance and causes runtime warnings.

## before fixed
微信图片_20260210121223_4_377
微信图片_20260210121223_5_377

## after fixed
image

## Release Notes

* **Documentation**
  * Corrected code examples in the State Management documentation to improve accuracy and consistency of API usage patterns.

### DIFF
--- a/packages/docs/zh/core-concepts/state.md
+++ b/packages/docs/zh/core-concepts/state.md
@@ -151,7 +151,7 @@ export default {
   computed: {
     // 可以访问组件中的 this.count
     // 与从 store.count 中读取的数据相同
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks!

PR #3115, opened 2026-03-31
feat: add a context API for stores

Ragnar-Oock:
Myself and others (see #1611 ) have had the need to provide things to stores during the initialization process, be it configuration or dependencies to be used in the store implementation without using global variables.

This is intended to be a first draft of a new API, I don't expect this PR to be merged as is.

## Implementation

I'm proposing the following implementation to add this capability.

I didn't use the Plugin API because plugins are applied after stores are created, making it impossible to inject anything in the setup/state function.

API modifications :

- added a `.provide()` method on the Pinia instance
- added a `context` property on the `SetupStoreHelpers` object
- added a `context` argument to the `state` function of Option stores
- added a `PiniaSetupContext` interface

## Subtasks

- [x] add a test suite for the new API
- [x] add a page of documentation with examples
- [ ] add link to the documentation page (I'm not too sure where to place it)

## Considerations and doubts :

1. I made the properties on the context object read-only when defined via the `.provide()` function to prevent mutations of the shared context from within the stores, so it's possible to add new properties but not to edit existing ones. I decided not to use `Object.freeze` because it would require unfreezing at the start of each call to `.provide()`, which might have adverse effects on performance.
2. I print a warning message in DEV when calling `.provide()` multiple times with the same key and abort the edition, but I'm not sure if the value should not be overridden instead.
3. I could not get the pre-commit hook to work on my machine, so I ran the formatting command manually. Feel free to point out any automated action that I inadvertently skipped by doing so. (I can also rewrite the history to split my edits and the formatting changes if needed)

&nbsp;

## Summary by CodeRabbit

- **New Features**
  - Added `pinia.provide()` method to inject initialization-time data and functions into stores
  - Context is accessible in both setup and options API store definitions
  - Support for TypeScript typing via module augmentation for custom context fields
- **Documentation**
  - Added guide explaining context injection with examples for typing and usage patterns

&nbsp;

### DIFF
--- a/packages/docs/core-concepts/store-context.md
+++ b/packages/docs/core-concepts/store-context.md
@@ -1,0 +1,141 @@
+# Passing context to stores
+
+<MasteringPiniaLink
+href="https://masteringpinia.com/lessons/What-is-a-pinia-plugin"
+title="Learn all about Pinia plugins"
+/>
+
+When building complex apps it can happen that you need to inject data or function
+in stores when they are first instanced.
+
+Context is added to the pinia instance with `pinia.provide()`. With it you can
+define a property of the context object that will be passed to the store when it
+is initialized.
+
+When using the setup API :
+
+```js
+import { createPinia, defineStore } from 'pinia'
+
+// add a property named `secret` to every store that is created
+// after this plugin is installed this could be in a different file
+const API_URL = 'https://localhost/blog/'
+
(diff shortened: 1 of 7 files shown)


posva:
Thanks for the PR but we don’t need this as it can be done with app.provide and inject already

the discussion you links to ask for something else btw

PR #3134, opened 2026-06-18
feat: gracefully skip nullish values in storeToRefs

### LINKED ISSUE #3123: storeToRefs throws TypeError: Cannot read properties of null (reading 'effect') when store state is null (Vue 3.5)

1. Use Pinia 3.0.4 with Vue 3.5+
2. Have an options store with any state property that starts as or is set back to null
3. Call an action that writes null to that property through the reactive proxy (e.g. a reset action)
4. Call storeToRefs on the same store instance in a different component/layout that mounts afterwards

### Expected behavior

storeToRefs returns refs for all reactive state properties and skips non-ref values (including null) gracefully.

### Actual behavior

TypeError: Cannot read properties of null (reading 'effect')
    at storeToRefs (pinia.mjs:1920)

<hr>

The crash occurs in storeToRefs at this line across all dist variants:

```js
// pinia.mjs ~line 1920
const value = rawStore[key];
if (value.effect) {  // ← no null guard; throws when value === null
```

### Additional information

Additional information

This bug exists in the published 3.0.4 on npm. The Pinia playground doesn't reproduce it, which suggests the fix is already in your source tree but hasn't been released. Could you publish a patch release (3.0.5) so users don't need workarounds like patch-package?

- Pinia version: 3.0.4
- Vue version: 3.5.x
- This is a Vue 3.5 regression. In Vue ≤ 3.4, writing `null` to reactive state through a Pinia store proxy kept the `ref()` wrapper intact in the store, so value was always a ref object (with effect at worst `undefined`). In Vue 3.5, the plain null is stored directly in the store, so `value.effect` throws.
- **The one-line fix is a null guard: `if (value && value.effect)`. We are currently working around this with patch-package on our end, but that is not a neat fix.**

JSap0914:
## Bug

`storeToRefs` throws `TypeError: Cannot read properties of null (reading 'effect')` when a setup store returns a non-ref `null` value as part of its state (closes #3123).

```ts
const useStore = defineStore('example', () => {
  return {
    // non-ref null is valid for optional state
    nullableItem: null as { id: number } | null,
    text: ref('hello'),
  }
})
// Before this fix: TypeError: Cannot read properties of null (reading 'effect')
const { text } = storeToRefs(useStore())
```

The loop in `storeToRefs` iterates all enumerable keys in the raw store. When a setup store returns a raw `null` value (not wrapped in `ref()`), `rawStore[key]` is literally `null`, and the unguarded `value.effect` access throws.

## Fix

Add a null-guard before the `.effect` access:

```diff
- if (value.effect) {
+ if (value && value.effect) {
```

The guard is consistent with how `isRef` and `isReactive` already handle nullish values. A null (or undefined) raw value is simply skipped — the same outcome as any other non-ref, non-reactive value.

## Verification

```
pnpm vitest run --project pinia packages/pinia/__tests__/storeToRefs.spec.ts
# → Tests 10 passed (10)
```

Red (before): test failed with `TypeError: Cannot read properties of null (reading 'effect')`
Green (after): all 10 storeToRefs tests pass, including the new regression test.

> *This PR was developed with AI assistance.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where `storeToRefs` could throw when a setup store includes optional state values that are `null` (non-ref).
  * Updated detection logic to safely handle nullish values before reading internal properties.

* **Tests**
  * Added coverage to ensure `storeToRefs` does not throw with `null` optional state and that existing refs are still returned and readable.

### DIFF
--- a/packages/pinia/src/storeToRefs.ts
+++ b/packages/pinia/src/storeToRefs.ts
@@ -94,7 +94,7 @@ export function storeToRefs<SS extends StoreGeneric>(
     const value = rawStore[key]
     // There is no native method to check for a computed
     // https://github.com/vuejs/core/pull/4165
-    if (value && value.effect) {
+    if (value?.effect) {
       // @ts-expect-error: too hard to type correctly
       refs[key] =
         // ...
--- a/packages/pinia/src/storeToRefs.ts
+++ b/packages/pinia/src/storeToRefs.ts
@@ -94,7 +94,7 @@ export function storeToRefs<SS extends StoreGeneric>(
     const value = rawStore[key]
     // There is no native method to check for a computed
     // https://github.com/vuejs/core/pull/4165
-    if (value.effect) {
+    if (value && value.effect) {
       // @ts-expect-error: too hard to type correctly
       refs[key] =
         // ...
--- a/packages/pinia/__tests__/storeToRefs.spec.ts
+++ b/packages/pinia/__tests__/storeToRefs.spec.ts
@@ -200,6 +200,23 @@ describe('storeToRefs', () => {
     expect(spy).toHaveBeenCalledTimes(0)
   })
 
+  it('does not crash when setup store returns null non-ref value', () => {
+    // A setup store may return a plain (non-ref) null value for optional state.
+    // storeToRefs must skip those gracefully instead of throwing
+    // TypeError: Cannot read properties of null (reading 'effect').
+    const useStore = defineStore('null-val', () => {
+      return {
+        nullableItem: null as null | { id: number },
+        text: ref('hello'),
+      }
+    })
+
+    const store = useStore()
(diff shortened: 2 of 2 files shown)


packages/pinia/src/storeToRefs.ts
```diff
@@ -94,7 +94,7 @@ export function storeToRefs<SS extends StoreGeneric>(
     const value = rawStore[key]
     // There is no native method to check for a computed
     // https://github.com/vuejs/core/pull/4165
-    if (value.effect) {
+    if (value && value.effect) {
```
posva:
```suggestion
    if (value?.effect) {
```

posva (approved):
Thanks! This is actually not valid but the runtime change is worth having.

PR #3138, opened 2026-06-20
fix: support thenables in action hooks

fallintoplace:
## Summary

- Treat promise-like action returns as async results for `$onAction` subscribers.
- Use `Promise.resolve()` to assimilate thenables before triggering `after` or `onError` hooks.
- Add coverage for native promises, custom thenables, rejected thenables, and cross-realm promises.

## Root cause

The action wrapper only checked `ret instanceof Promise`, so promise-like objects and promises created in another realm skipped the async hook path. That meant `after` could receive the unresolved object and `onError` could miss rejected thenables.

## Validation

- `pnpm vitest run packages/pinia/__tests__/onAction.spec.ts`
- `pnpm run test:types`
- `pnpm exec oxfmt --check packages/pinia/src/store.ts packages/pinia/__tests__/onAction.spec.ts`
- `pnpm run test:vitest run`

## Release Notes

* **Bug Fixes**
  * Improved action subscription handling to properly recognize and normalize promise-like return values, ensuring consistent behavior across different Promise implementations and contexts.

* **Tests**
  * Expanded test coverage for action return and rejection handling with custom thenables and cross-realm Promises.

### DIFF
--- a/packages/pinia/__tests__/onAction.spec.ts
+++ b/packages/pinia/__tests__/onAction.spec.ts
@@ -2,6 +2,7 @@ import { beforeEach, describe, it, expect, vi } from 'vitest'
 import { createPinia, defineStore, setActivePinia } from '../src'
 import { mount } from '@vue/test-utils'
 import { nextTick } from 'vue'
+import { createContext, runInContext } from 'node:vm'
 
 describe('Subscriptions', () => {
   const useStore = () => {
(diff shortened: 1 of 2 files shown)


posva:
What bug is it solving? that node js part doesn't look realistic and this PR looks like just AI generated

fallintoplace:
Real case would be iframe / embedded app code returning a Promise from another window. It’s awaitable, but instanceof Promise can be false, so $onAction treats it as sync and after() gets the Promise instead of the resolved value.

I can drop that test or add a comment to make the intent clearer.

posva:
What I need is an actual reproduction

posva:
Let me know if you can provide a real-world case scenario so we can reconsider this

Issue #3155, opened 2026-07-15
4.0.1 published to npm still has incorrect typing definition


kubajmarek:
### Reproduction

Install 4.0.1

### Steps to reproduce the bug

1. Install 4.0.1 or check https://registry.npmjs.org/pinia/4.0.1 or https://unpkg.com/pinia@4.0.1/package.json
2. Check for `.d.ts` file

### Expected behavior

`.d.ts` is in exports

### Actual behavior

`.d.ts` is not in exports

WuMingDao:
yes, I use create-vue create and install pinia 4.0.1 still not type export

<img width="1411" height="424" alt="Image" src="https://github.com/user-attachments/assets/60c8d8f1-7715-43e0-86af-ab206fe613a7" />

StillAlexLiu:
same issues

WuMingDao:
I fixed it, it is tsdown option question by #3156  😉

posva:
Fixed in #3158

PR #3156, opened 2026-07-15
fix: pinia@4.0.1 still not export of types

WuMingDao:
## Summar

It is tsdown option to it, exports set false, build have export of type

## I Found flow

By #3155

In pinia@4.0.1 still not export of type

image

image

but by #3154 fixed it, so not source code question, maybe bunder or npm pubic flow question?

I found use tsdown build pinia, package.json modify

image

So I think the question at tsdown

image

question is exports: true that option

image

because it modify package.json

image

set false, this ok

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility of the ESM build output with supported module environments.


sxzz:
I can fix it.

PR #3162, opened 2026-07-17
fix(store): replace reactive Set/Map state on hydration instead of unioning it

### LINKED ISSUE #3161: Reactive Set/Map state is unioned instead of replaced during SSR hydration
When a **setup store** has a `reactive(new Set([...]))` / `reactive(new Map([...]))` state property with **default entries**, SSR hydration **unions** the client defaults with the server state instead of **replacing** them, so entries the server removed survive on the client.

### Reproduction

```ts
const useStore = defineStore('main', () => {
  const setReactive = reactive(new Set(['default']))
  return { setReactive }
})

const pinia = createPinia()
pinia.state.value.main = { setReactive: new Set(['a']) }  // server state
setActivePinia(pinia)

const store = useStore()
console.log([...store.setReactive])   // ['default', 'a']  <- wrong; should be ['a']
```

The sibling `ref()` branch replaces correctly (`prop.value = initialState[key]`), and HMR replaces too — the server state is authoritative. The Set/Map case only diverges.

### Cause

The reactive-hydration branch calls `mergeReactiveObjects`, whose Set/Map branches do `patchToApply.forEach(target.add, target)` (union) without clearing the target first. The existing hydration tests use **empty** default collections, where union == replace, so this was never caught. It's a real Nuxt+Pinia scenario (devalue serializes Set/Map). PR incoming.

winklemad:
Fixes #3161.

### Problem
When a setup store's state has a `reactive(new Set([...]))` / `reactive(new Map([...]))` with default entries, SSR hydration **unions** the defaults with the server state instead of **replacing** them, so entries the server removed survive (repro in #3161). The sibling `ref()` branch and HMR both replace; the server state is authoritative.

### Cause
The reactive-hydration branch calls `mergeReactiveObjects`, whose Set/Map branches add server entries onto the client collection without clearing it first.

### Fix
Clear a reactive `Set`/`Map` before merging **in the hydration path only** (`store.ts`), leaving `$patch`'s own merge semantics (which also calls `mergeReactiveObjects`) unchanged.

### Tests
Added a hydration test with **non-empty** reactive Set/Map defaults asserting the server state replaces them (the existing tests used empty defaults, so union == replace and the bug was invisible). It fails before this change and passes after; the full pinia suite stays green (211 passed), including `$patch`, and typecheck is clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SSR hydration for reactive `Set` and `Map` values in setup stores.
  * Hydrated `Set`/`Map` contents now fully replace existing default entries, preventing stale items from lingering after hydration.
* **Tests**
  * Added coverage to verify `Set`/`Map` hydration behavior for both top-level and nested reactive structures.

### DIFF
--- a/packages/pinia/__tests__/state.spec.ts
+++ b/packages/pinia/__tests__/state.spec.ts
@@ -315,6 +315,26 @@ describe('State', () => {
     expect([...store.setReactive.values()]).toEqual([3, 4])
   })
 
+  it('replaces (does not union) a non-empty reactive Set/Map on hydration', async () => {
+    const useStore = defineStore('main', () => {
+      const setReactive = reactive(new Set(['default']))
+      const mapReactive = reactive(new Map([['x', 1]]))
+      return { setReactive, mapReactive }
+    })
+
+    const pinia = createPinia()
+    pinia.state.value.main = {
+      setReactive: new Set(['a']),
+      mapReactive: new Map([['y', 2]]),
+    }
+    setActivePinia(pinia)
+
+    const store = useStore()
+    // The server state is authoritative, so the default entries are gone.
+    expect([...store.setReactive.values()]).toEqual(['a'])
+    expect([...store.mapReactive.entries()]).toEqual([['y', 2]])
+  })
+
   it('hydrates Map in option stores', async () => {
     const useStore = defineStore('main', {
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

PR #3168, opened 2026-07-25
fix: guard __VUE_PROD_DEVTOOLS__ in dist output

### LINKED ISSUE #3167: Pinia 4 + nuxt: fail on building pre-rendered pages
### Reproduction

https://stackblitz.com/edit/github-jm6ehxba?file=package.json

Hi 👋 

After updating to @pinia/nuxt 1.0 from version 0.11 and also from pinia 3 to 4 I saw that `pnpm run build` started to fail on pre-rendered pages.

The reproduction repo has:
- Nuxt 4.5
- Pinia 4 + @pinia/nuxt 1.0
- Nitro pre-rendering on "/" page

## To reproduce

1. Run `npm run build` with `pinia^4` + `@pinia/nuxt^1`
2. Run `npm run build`, it fails with access to undefined state
3. Downgrade to `pinia^3` + `@pinia/nuxt^0.11`
4. Run `npm run build` to see the build passing fine

If you remove the prerendering from the route on `nuxt.config.ts` everything works like in `v.0.11`

### Expected behavior

While running on Pinia 3, Nitro pre-rendering worked fine on Build, this is expected on Pinia 4 too

### Actual behavior

While building a pre-rendered route using Pinia 4 the build process fails with a mention on:

`Cannot read property of undefined (reading state)`

### Additional information

_No response_

ShreeBohara:
close #3167

**What was broken**

Since the tsdown build migration (898b7c9, #3092), the `__USE_DEVTOOLS__` define inlines `__VUE_PROD_DEVTOOLS__` as a bare global into `dist/pinia.js`:

```js
if ((process.env.NODE_ENV !== "production" || __VUE_PROD_DEVTOOLS__) && ...
```

The v3 rollup build emitted a guarded form instead: `typeof __VUE_PROD_DEVTOOLS__ !== 'undefined' && __VUE_PROD_DEVTOOLS__`. Without the guard, evaluating the dist file anywhere a bundler hasn't defined the global throws `ReferenceError: __VUE_PROD_DEVTOOLS__ is not defined` inside `createPinia()`.

That is what #3167 actually hits: Nitro's prerenderer externalizes `node_modules`, so its chunks import `dist/pinia.js` directly in Node. `createPinia()` throws, the `@pinia/nuxt` plugin `setup()` fails, `$pinia` is never provided, and the visible symptom becomes the misleading `Cannot read properties of undefined (reading 'state')` from the `app:rendered` hook. The same ReferenceError reproduces without Nuxt in any plain-Node consumption of the dist file (SSR with externalized deps, node test setups without the vite plugin's defines).

**Reproduction**

Bare Node, pinia 4.0.2:

```
NODE_ENV=production node --input-type=module \
  -e "const { createPinia } = await import('pinia'); createPinia()"
```

throws the ReferenceError; with this branch it runs. Also verified end to end with a minimal Nuxt 4.5 + @pinia/nuxt 1.0.1 app with one prerendered route (matching the issue's reproduction): `nuxt build` fails on 4.0.2 and succeeds with a pack of this branch, with `@pinia/nuxt` unchanged. The same app on pinia 3 / @pinia/nuxt 0.11 builds fine, which brackets the regression.

**The change**

One line in `packages/pinia/tsdown.config.ts`: restore the typeof guard inside the `__USE_DEVTOOLS__` define, so the dist output degrades to `false` when the global is not defined. Bundlers that do define `__VUE_PROD_DEVTOOLS__` constant-fold the expression exactly as before.

Alternatives considered: guarding each call site in `src` (more invasive, same emitted result), or changing `@pinia/nuxt` to not rely on injections in `app:rendered` (does not fix the underlying ReferenceError, which reproduces without Nuxt).

**Testing**

Full local suite is green (lint, types, vitest, builds, dts). There is no in-repo test that exercises built dist output, so per the CONTRIBUTING note about hard-to-test changes: I am happy to add a small post-build smoke script (import `dist/pinia.js` in Node with `NODE_ENV=production` and call `createPinia()`) if you want one — tell me where you would like it wired.

Disclosure: this change was investigated and prepared with AI assistance (Claude Code). I reviewed it and can explain every line.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility for direct Node.js imports and environments without bundler-defined production flags.
  - Preserved existing development and test-mode behavior for devtools availability.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved direct Node.js imports of the production build when bundler configuration is unavailable.
  * Preserved existing development and test-mode behavior.

* **Tests**
  * Added coverage for importing the production build directly and verifying store actions work correctly.
  * Added distribution tests to local and continuous integration test runs.

* **Chores**
  * Updated test commands to include distribution import validation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
       - run: pnpm run test:vitest
       - run: pnpm run build
       - run: pnpm run test:dts
+      - run: pnpm run test:dist
       - run: pnpm run size
 
       - uses: codecov/codecov-action@v5
--- a/package.json
+++ b/package.json
@@ -21,10 +21,11 @@
     "lint": "oxfmt --check",
     "fmt": "oxfmt",
     "lint:fix": "pnpm fmt",
-    "test": "pnpm run -r dev:prepare && pnpm run test:types && pnpm run test:vitest run && pnpm run -r test && pnpm run build && pnpm test:dts",
+    "test": "pnpm run -r dev:prepare && pnpm run test:types && pnpm run test:vitest run && pnpm run -r test && pnpm run build && pnpm test:dts && pnpm test:dist",
     "test:vitest": "vitest --coverage",
     "test:types": "tsc --build ./tsconfig.json",
     "test:dts": "pnpm run -r test:dts",
+    "test:dist": "pnpm run -r test:dist",
     "postinstall": "simple-git-hooks"
   },
   "devDependencies": {
--- a/packages/pinia/__tests__/dist-import.mjs
+++ b/packages/pinia/__tests__/dist-import.mjs
@@ -1,8 +1,8 @@
 // @ts-check
 // `dist/pinia.js` must run in plain Node, where no bundler defines
 // `__VUE_PROD_DEVTOOLS__`: unguarded, `__USE_DEVTOOLS__` throws a
-// ReferenceError inside `createPinia()` (#3167). Needs a built `dist/`, so it
-// runs after the build, as part of `test:dts`.
+// ReferenceError inside `createPinia()` (#3167). Needs a built `dist/`, so
+// `test:dist` only runs after `build`.
 
 // set before importing: with NODE_ENV=production `__DEV__` is false, so the
 // `__VUE_PROD_DEVTOOLS__` side of `__USE_DEVTOOLS__` is actually evaluated.
--- a/packages/pinia/__tests__/dist-import.mjs
+++ b/packages/pinia/__tests__/dist-import.mjs
@@ -1,18 +1,20 @@
 // @ts-check
-// Ensures the built `dist/pinia.js` works when imported directly in Node
-// without a bundler injecting compile-time defines (#3167): if
-// `__VUE_PROD_DEVTOOLS__` is left unguarded in the dist output, evaluating
-// `__USE_DEVTOOLS__` throws a ReferenceError. Requires `dist/` to be built,
-// so it runs after the build, as part of `test:dts`.
-import { spawnSync } from 'node:child_process'
-import { fileURLToPath } from 'node:url'
+// `dist/pinia.js` must run in plain Node, where no bundler defines
+// `__VUE_PROD_DEVTOOLS__`: unguarded, `__USE_DEVTOOLS__` throws a
+// ReferenceError inside `createPinia()` (#3167). Needs a built `dist/`, so it
+// runs after the build, as part of `test:dts`.
 
-const script = `
-import { createApp } from 'vue'
-import { createPinia, defineStore } from './dist/pinia.js'
+// set before importing: with NODE_ENV=production `__DEV__` is false, so the
+// `__VUE_PROD_DEVTOOLS__` side of `__USE_DEVTOOLS__` is actually evaluated.
+// static imports are hoisted, hence the dynamic ones below
+process.env.NODE_ENV = 'production'
+
+const { createApp } = await import('vue')
+const { createPinia, defineStore } = await import('../dist/pinia.js')
 
 const pinia = createPinia()
 createApp({}).use(pinia)
+
 const useCounterStore = defineStore('counter', {
   state: () => ({ n: 0 }),
   actions: {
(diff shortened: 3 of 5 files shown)


posva:
im not sure we can go this way because it probably doesn’t tree shake

kevinmarrec:
<details>
<summary>Tree-shaking measurements (written by Claude Code, run locally by me)</summary>

`define` **is** substituted inside `typeof` — verified with esbuild, Vite 7, Vite 8/rolldown, webpack `DefinePlugin` and `@rollup/plugin-replace`. It folds in four steps before tree-shaking runs:

`typeof false !== 'undefined' && false` → `'boolean' !== 'undefined' && false` → `true && false` → `false`

Built `main` and this branch from source, then bundled a production Vue app against each dist:

| | `main` (bare) | this PR |
|---|---|---|
| Vite 8 prod app | 89,134 B | 89,134 B — identical content hash |
| Vite 7 prod app | 84,833 B | 84,833 B — identical content hash |
| Node, `NODE_ENV=production` | `ReferenceError` | OK |

Zero devtools payload in both bundles (no 🍍 strings, no `setupDevtoolsPlugin`, no `@vue/devtools-api`). The guard is self-cancelling: it only matters against an undeclared global at runtime, which is exactly the case where no `define` ran — so it can't cost tree-shaking and save Node in the same build.

</details>

@ShreeBohara it'd be worth reproducing these numbers on your side before @posva has to take them on trust — build Pinia on `main` and on your branch, bundle the same app against each, compare.

ShreeBohara:
I tried to verify this rather than argue it. Built `v4` and this branch and measured consumer bundles for a minimal app (createApp + createPinia + one setup store), production mode, with `__VUE_PROD_DEVTOOLS__: false` as @vitejs/plugin-vue injects it:

| toolchain | v4 | PR | delta |
|---|---|---|---|
| Vite 7.3.6 + plugin-vue 6.0.8 | 65,946 | 65,946 | byte-identical |
| webpack 5.109.0 + DefinePlugin | 65,827 | 65,827 | byte-identical |
| rollup 4 + plugin-replace + terser | 64,090 | 64,090 | byte-identical |
| esbuild 0.28.1 --define | 68,377 | 68,377 | same size, differs only in mangled identifier order |

Zero occurrences of `setupDevtoolsPlugin`, `@vue/devtools-api`, or `__VUE_PROD_DEVTOOLS__` in any of the eight bundles. Positive control: rebuilt with the flag defined as `true`, Vite goes to 136,821 (+70,875), so the measurement does detect DCE failure. Also checked esbuild, rolldown, rspack, bun and terser's `global_defs` separately — all substitute inside `typeof` and fold.

Cost is +240 bytes raw / +6 gzip in `dist/pinia.js` itself. The other dist artifacts are byte-identical, since `esm-browser.prod` and `iife.prod` hardcode `__USE_DEVTOOLS__`.

Two things worth flagging against the change:

1. If a consumer's bundler defines nothing at all (Parcel, bare rollup), the PR adds ~200 bytes and both versions ship the devtools code anyway. The guard stops the crash there, it doesn't restore tree-shaking.
2. Vue's `initFeatureFlags` assigns `getGlobalThis().__VUE_PROD_DEVTOOLS__ = false`, so with no define, calling `createApp()` before any pinia code already avoids the ReferenceError. The crash in #3167 needs pinia to run first — module-scope store use, SSR, tests.

If you'd rather not carry `typeof` in the DCE gate at all, the alternative is Vue's own approach: set the global once in an init path and keep the bare identifier at the use sites. Happy to redo it that way, or to close this if you consider the no-define case out of scope.

kevinmarrec:
@posva The fix looks safe and enable (fix) the Pinia v4 upgrade for multiple `Vite + preprender` based frameworks (Nuxt.js and Vike confirmed, but maybe also Quasar / VitePress / etc...)

Let us know if you need any additional context to make this land :)

@ShreeBohara We could maybe add some testing against this change if relevant

ShreeBohara:
Added in b0499d0d: `__tests__/dist-import.mjs` spawns a bare Node child — no bundler, no compile-time defines — that imports the built `dist/pinia.js`, installs `createPinia()` and uses a store, i.e. the #3167 scenario. It runs under `NODE_ENV=production` (the define is inlined lazily and `__DEV__` short-circuits it otherwise, so production is the leg that actually evaluates `__VUE_PROD_DEVTOOLS__`), plus once with `NODE_ENV` unset as cheap insurance in case the define ever becomes eagerly evaluated.

Reverting the guard in `tsdown.config.ts` and rebuilding makes it fail with `ReferenceError: __VUE_PROD_DEVTOOLS__ is not defined at createPinia`, and the failure propagates through the root `test:dts`, so CI goes red.

On placement: CI runs the vitest step before `build`, so a vitest spec needing `dist/` would never actually run in CI. `test:dts` is the existing post-build check, so the script is wired as a separate `test:dist` chained from there — a 2-line package.json change, no ci.yml edit. If a dedicated CI step or another location is preferred, the test file needs no changes, only re-wiring.

Scope note: only `dist/pinia.js` references `__VUE_PROD_DEVTOOLS__` (the browser/iife builds hardcode `__USE_DEVTOOLS__`), and it's the sole `exports` entry. The `.mjs` isn't collected by vitest and isn't published to npm.

posva (approved):
Thanks a lot!

PR #3178, opened 2026-09-01
fix: keep notifying subscriptions when $patch throws

giaBaoJS:
`$patch()` sets `isListening` and `isSyncListening` to `false`, applies the mutation, then sets them back. The restore is not protected against the mutation throwing, so if the mutator (or the object merge) throws, the flags stay `false` and every `$subscribe()` callback of that store silently stops firing on direct mutations. It only recovers if a later `$patch()` happens to succeed.

`packages/pinia/src/store.ts`, `$patch()`: the assignments after the `if/else` are skipped when the mutation throws.

### Reproduction

```ts
const store = useStore()
store.$subscribe(spy, { flush: 'sync' })

try {
  store.$patch((state) => {
    state.user = 'Ana'
    throw new Error('failed patch')
  })
} catch {}

store.user = 'Cleiton'
// spy was never called, and never will be
```

### Fix

Move the restore into a `finally` block. The error still propagates and `triggerSubscriptions()` is still skipped for the failed patch, so nothing is reported for a mutation that did not complete, but the store keeps notifying afterwards.

### Test

Added to the existing `Options Stores` / `Setup Stores` matrix in `subscriptions.spec.ts`, covering both store flavours and both the `sync` and the default flush. It fails on `v4` and passes with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where store subscriptions could stop responding after a state patch operation threw an error.
  * Subscriptions now continue receiving notifications after failed patch callbacks or state merges.
  * Patch errors continue to be reported while subscription behavior remains consistent.

### DIFF
--- a/packages/pinia/__tests__/subscriptions.spec.ts
+++ b/packages/pinia/__tests__/subscriptions.spec.ts
@@ -181,6 +181,27 @@ describe('Subscriptions', () => {
       expect(spySync).toHaveBeenCalledTimes(2)
     })
 
+    it('keeps notifying subscribers if $patch throws', async () => {
+      const store = useStore()
+      const spySync = vi.fn()
+      const spy = vi.fn()
+      store.$subscribe(spySync, { flush: 'sync' })
+      store.$subscribe(spy)
+
+      expect(() =>
+        store.$patch((state): void => {
+          state.user = 'Ana'
+          throw new Error('failed patch')
+        })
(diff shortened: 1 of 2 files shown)


posva:
Thanks but your patch mutator function should never throw: it's user responsibility to apply a try/catch in the parts where an exception can be thrown

PR #3179, opened 2026-09-02
fix: replace reactive arrays on hydration instead of merging into them

contactjawad:
With a setup store whose state has a reactive array, hydration leaves stale default items behind when the incoming state is shorter:

```js
// store: () => ({ list: reactive(['a', 'b', 'c']) })
// hydrated with { list: ['x'] }
store.list // ['x', 'b', 'c'], expected ['x']
```

On hydration the code clears `Set`/`Map` before merging so the server value wins, but a reactive array isn't a `Set`/`Map`, so `mergeReactiveObjects` just assigns by index and never shortens it. A `ref([...])` array is fine because it's replaced wholesale.

I added an `Array.isArray(prop)` case next to the Set/Map clear that resets the length before the merge, so the array is replaced the same way. Added a test (with a `ref` control that already passed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed state hydration so server-provided array values replace default array contents instead of being appended.
  * Ensured reactive arrays and array references are cleared before hydrated state is applied.

### DIFF
--- a/packages/pinia/__tests__/state.spec.ts
+++ b/packages/pinia/__tests__/state.spec.ts
@@ -335,6 +335,26 @@ describe('State', () => {
     expect([...store.mapReactive.entries()]).toEqual([['y', 2]])
   })
 
+  it('truncates a reactive Array on hydration', async () => {
+    const useStore = defineStore('main', () => {
+      const listReactive = reactive(['a', 'b', 'c'])
+      const listRef = ref(['a', 'b', 'c'])
+      return { listReactive, listRef }
+    })
+
+    const pinia = createPinia()
+    pinia.state.value.main = {
+      listReactive: ['x'],
+      listRef: ['x'],
+    }
+    setActivePinia(pinia)
+
+    const store = useStore()
+    // The server state is authoritative, so the default entries are gone.
+    expect([...store.listReactive]).toEqual(['x'])
+    expect([...store.listRef]).toEqual(['x'])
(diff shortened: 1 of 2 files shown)


lvyanyan (commented):
Reviewed on top of v4 (a4eb2c5). The change is correct and well-localized — the truncation happens at the hydration call site in `createSetupStore`, so `$patch` keeps its index-merging semantics (same reason the existing `Set`/`Map` `clear()` lives there). I verified a longer server array still fills correctly, and arrays nested inside reactive objects are replaced wholesale through the object-recursion path, so they never hit the stale-merge problem. The full pinia suite passes on the PR.

One edge case worth an explicit decision: when the incoming server state does not contain the key at all (or contains `undefined`), the array is now wiped where it previously kept its default:

```js
const useStore = defineStore('main', () => ({ list: reactive(['a', 'b', 'c']) }))
pinia.state.value.main = {} // server payload without `list`
const store = useStore()
store.list // now `[]`, was `['a', 'b', 'c']`
```

This matches the existing `Set`/`Map` `clear()` behavior (which also wipes when the key is missing), but arrays are far more common as state defaults, so the impact surface is bigger. If you want to keep defaults for missing keys, gating on `Array.isArray(initialState[key])` before truncating would do it. Either way is defensible — flagging so it's a conscious choice, and a test for the chosen behavior would be welcome.

PR #3180, opened 2026-09-02
fix: include plugin-added reactive() state in storeToRefs

contactjawad:
If a plugin adds a `reactive()` object to the store, `storeToRefs()` silently drops it:

```js
pinia.use(() => ({ shared: reactive({ x: 1 }) }))
storeToRefs(store) // no `shared`
```

Plugin extensions are applied with `assign(store, extensions)`, which writes through the reactive proxy, so the raw target ends up with the unwrapped (plain) value. `storeToRefs` iterates `toRaw(store)` and skips anything that isn't reactive, so the plugin's reactive state doesn't make it. Setup-store state avoids this because `createSetupStore` also assigns onto `toRaw(store)`; plugin extensions never got that.

I added the same `assign(toRaw(store), extensions)` right after the existing assign, so the reactive proxy is present on the raw target. A `ref` from a plugin already worked; this makes `reactive()` behave the same. Added a test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin-provided reactive objects are now correctly exposed by `storeToRefs()`.
  * Updates to reactive objects added by plugins are reflected through their extracted refs.

* **Tests**
  * Added coverage verifying reactive plugin state behavior with `storeToRefs()`.

### DIFF
--- a/packages/pinia/__tests__/storeToRefs.spec.ts
+++ b/packages/pinia/__tests__/storeToRefs.spec.ts
@@ -153,6 +153,28 @@ describe('storeToRefs', () => {
     ).toEqual(objectOfRefs({ n: 0, pluginN: 20 }))
   })
 
+  it('contain plugin reactive() states', () => {
+    const pinia = createPinia()
+    // directly push because no app
+    pinia._p.push(() => ({
+      // @ts-expect-error: cannot set a ref yet
+      pluginN: ref(20),
+      reactiveN: reactive({ x: 1 }),
+    }))
+    setActivePinia(pinia)
+
+    const store = defineStore('a', {
+      state: () => ({ n: 0 }),
+    })()
+
+    const refs = storeToRefs(store)
+    expect(refs).toHaveProperty('reactiveN')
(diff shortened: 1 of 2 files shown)


lvyanyan (commented):
Verified this on top of v4 (a4eb2c5) — the red/green is real: the added test fails on the base branch and passes with the fix. With the change, `toRaw(store)` holds the reactive proxy, `storeToRefs()` picks it up, mutations flow through the extracted value, and nothing leaks into `$state`. The full pinia suite passes.

One non-blocking scope note: the fix covers values present at registration time. A later whole-property reassignment goes back through the reactive proxy's setter, which normalizes reactive values to raw, so the key silently disappears from `storeToRefs()` again:

```js
store.shared = reactive({ y: 2 })
storeToRefs(store) // no `shared` anymore
```

Same root cause as the original report, just at a different point in time. Probably acceptable for this PR — fixing it would require `storeToRefs` to stop iterating `toRaw(store)` for this case, a bigger change — but worth being explicit about.

